Issue #12, opened 2023-05-19
libbpf: map 'simple_ops': failed to create: Operation not supported(-95)


EdLovecraft:
```
❯ sudo ./scx_example_simple
libbpf: map 'simple_ops': failed to create: Operation not supported(-95)
libbpf: failed to load object 'scx_example_simple'
libbpf: failed to load BPF skeleton 'scx_example_simple': -95
scx_example_simple: scx_example_simple.c:83: int main(int, char **): Assertion `!scx_example_simple__load(skel)' failed.
[1]    26428 IOT instruction  sudo ./scx_example_simple
```

ByteLab-David:
@anakryiko this seems to be a fallout from https://github.com/sched-ext/sched_ext/commit/0db0ad6defeacfc957cc3f7394a17870bf1ad16a. I tested it locally and it seemed to work, but maybe I made a mistake because this is now happening to me as well. Did we need any commits for this beyond https://github.com/sched-ext/sched_ext/commit/8d1608d70927747da9c1a8770edf7b6ee68f8ebc?

ByteLab-David:
Let me revert 0db0ad6defeacfc957cc3f7394a17870bf1ad16a for now and then we can continue figuring out what the issue was.

ByteLab-David:
@EdLovecraft can you try pulling the latest commit again and rebuilding the schedulers, and seeing if that fixes it for you?

EdLovecraft:
> @EdLovecraft can you try pulling the latest commit again and rebuilding the schedulers, and seeing if that fixes it for you?

@Decave  Fixed, thank you

ByteLab-David:
I'm going to try rebasing on top of the latest bpf-next master branch to see if that fixes it. That seems useful regardless in preparation for v4 of the patch set. Unfortunately, it seems that bpf-next master fails to boot at the moment, so it might take some time today to figure this out.

ByteLab-David:
> I'm going to try rebasing on top of the latest bpf-next master branch to see if that fixes it. That seems useful regardless in preparation for v4 of the patch set. Unfortunately, it seems that bpf-next master fails to boot at the moment, so it might take some time today to figure this out.

[...]

PR #14, opened 2023-05-22
Add proper support for auto-detach in sched_ext

Decave:
When a BPF scheduler user-space program exits, the struct_ops map that                                                                                                                        
it attaches to to load the BPF program is not automatically detached.                                                                                                                         
This can cause the map and progs to remain loaded, which in turn will                                                                                                                         
cause future attempts to load a new scheduling program to fail.                                                                                                                               
                                                                                                                                                                                              
In commit 8d1608d70927 ("libbpf: Create a bpf_link in                                                                                                                                         
bpf_map__attach_struct_ops()"), Kui-feng updated                                                                                                                                              
bpf_map__attach_struct_ops() to create an actual link for the struct_ops                                                                                                                      
map, which in turn makes it automatically close the map when the owning                                                                                                                       
program exits. This patch updates the example schedulers to leverage                                                                                                                          
this by updating their sections to SEC(".struct_ops.link").

When I originally tried to do this, I neglected to actually update the core
ext.c code to support linked struct_ops. This set now includes the necessary
changes to those files, so #12 shouldn't happen again (and I tested all of
the schedulers to verify that it does not).

### DIFF
--- a/kernel/sched/ext.c
+++ b/kernel/sched/ext.c
@@ -3555,6 +3555,23 @@ static int bpf_scx_init(struct btf *btf)
 	return 0;
 }
 
+static int bpf_scx_update(void *kdata, void *old_kdata)
+{
+	/*
+	 * sched_ext does not support updating the actively-loaded BPF
+	 * scheduler, as registering a BPF scheduler can always fail if the
+	 * scheduler returns an error code for e.g. ops.init(),
+	 * ops.prep_enable(), etc. Similarly, we can always race with
+	 * unregistration happening elsewhere, such as with sysrq.
+	 */
+	return -EOPNOTSUPP;
+}
+
+static int bpf_scx_validate(void *kdata)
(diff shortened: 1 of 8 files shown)


anakryiko (approved):
LGTM. Consider following up with patches to bpf-next to make update/validate optional, and Alexei's suggested idea to always enforcing BPF_F_LINK flag for sched_ext.

ByteLab-David:
Thanks @anakryiko, will do.

>Alexei's suggested idea to always enforcing BPF_F_LINK flag for sched_ext.

I didn't actually see anywhere that we pass the map flags to the implementer of the bpf_struct_ops structure, or I would have also included that in the patch set. I think that will also require some changes in bpf-next.

anakryiko:
yep, seems like you'd need to extend bpf_struct_ops with a new callback to enforce this

PR #16, opened 2023-06-04
sched_ext: Fix code comment as scx_bpf_switch_all doesn't require 'into_scx'

shun159:


kernel/sched/ext.c
```diff
@@ -3702,11 +3702,10 @@ __diag_ignore_all("-Wmissing-prototypes",
 
 /**
  * scx_bpf_switch_all - Switch all tasks into SCX
- * @into_scx: switch direction
  *
- * If @into_scx is %true, all existing and future non-dl/rt tasks are switched
- * to SCX. If %false, only tasks which have %SCHED_EXT explicitly set are put on
- * SCX. The actual switching is asynchronous. Can be called from ops.init().
+ * Switch all existing and future non-dl/rt tasks to SCX.
+ * This can be called from ops.init() and actual switching
```
ByteLab-David: Suggestion: Maybe it would be slightly clearer to say, "This can only be called from ops.init(), and the actual switching is performed asynchronously."

What do you think?

ByteLab-David:
Thanks for catching this. Looks good, just had a suggestion to slightly improve verbiage

kernel/sched/ext.c
shun159: Yes, I totally agree with you. I fixed the comment as your suggestion. :pray:

shun159:
Thank you for merging!

ByteLab-David:
> Thank you for merging!

Of course, thank you for your contributions! Please keep em' coming!

PR #18, opened 2023-06-10
scx: Remove unnecessary pr_err on ops.prep_enable() failure path

Decave:
When ops.prep_enable() returns an error code that is propagated through scx_ops_prepare_task(), we currently issue a pr_err from scx_ops_enable(). It seems wrong for a BPF prog to be able to cause the kernel to issue a pr_err message, and we already capture the error in an ops_sanitizer_err() call in scx_ops_prepare_task(). Let's remove it.


ByteLab-David:
Tabling this for now -- we don't print the task name on the ops_sanitize_err() path. We can improve sanitize error messaging and generate a custom message in the init path later.

PR #20, opened 2023-06-15
use resizing of datasec maps in examples

inwardvessel:
there are some percpu arrays in the example schedulers that can make use of a recent libbpf feature that allows for resizing global data sections. for select arrays, they can be placed in custom named data sections and then resized to match the actual cpu count on a given host.

shown below is the BTF dump of the central and pair example scheduler programs that shows the array lengths have changed (nr_elements) to match the 12-cpu host they were tested on. note the linkage has also changed. this is a result of making the arrays non-static which was needed to access them through the skeleton for resizing in the userspace program.

central_dispatch before:
[742] ARRAY '(anon)' type_id=97 index_type_id=4 nr_elems=4096
[743] VAR 'cpu_gimme_task' type_id=742, linkage=static
[744] ARRAY '(anon)' type_id=53 index_type_id=4 nr_elems=4096
[745] VAR 'cpu_started_at' type_id=744, linkage=static

central_dispatch after:
[743] VAR 'cpu_gimme_task' type_id=867, linkage=global
[745] VAR 'cpu_started_at' type_id=868, linkage=global
[867] ARRAY '(anon)' type_id=97 index_type_id=4 nr_elems=12
[868] ARRAY '(anon)' type_id=53 index_type_id=4 nr_elems=12

pair_dispatch before:
[675] ARRAY '(anon)' type_id=673 index_type_id=4 nr_elems=4096
[676] VAR 'pair_cpu' type_id=675, linkage=global
[733] ARRAY '(anon)' type_id=669 index_type_id=4 nr_elems=4096
[734] VAR 'pair_id' type_id=733, linkage=global
[735] VAR 'in_pair_idx' type_id=733, linkage=global

pair_dispatch after:
[730] VAR 'pair_cpu' type_id=890, linkage=global
[732] VAR 'pair_id' type_id=891, linkage=global
[733] VAR 'in_pair_idx' type_id=892, linkage=global
[890] ARRAY '(anon)' type_id=727 index_type_id=4 nr_elems=12
[891] ARRAY '(anon)' type_id=667 index_type_id=4 nr_elems=12
[892] ARRAY '(anon)' type_id=667 index_type_id=4 nr_elems=12

the existing MEMBER_VPTR() macro was no longer applicable to use on these select arrays, because it uses the compile-time length so it caused verification failures when attempting to access an array element at an index greater than the initial size. a variation on this macro was added that accepts a size argument in order to match the newly resized arrays.

### DIFF
--- a/tools/sched_ext/scx_central.bpf.c
+++ b/tools/sched_ext/scx_central.bpf.c
@@ -51,14 +51,13 @@ char _license[] SEC("license") = "GPL";
 
 enum {
 	FALLBACK_DSQ_ID		= 0,
-	MAX_CPUS		= 4096,
 	MS_TO_NS		= 1000LLU * 1000,
 	TIMER_INTERVAL_NS	= 1 * MS_TO_NS,
 };
 
 const volatile bool switch_partial;
 const volatile s32 central_cpu;
-const volatile u32 nr_cpu_ids = 64;	/* !0 for veristat, set during init */
+const volatile u32 nr_cpu_ids = 1;	/* !0 for veristat, set during init */
 
 u64 nr_total, nr_locals, nr_queued, nr_lost_pids;
 u64 nr_timers, nr_dispatches, nr_mismatches, nr_retries;
@@ -73,8 +72,8 @@ struct {
 } central_q SEC(".maps");
 
 /* can't use percpu map due to bad lookups */
-static bool cpu_gimme_task[MAX_CPUS];
-static u64 cpu_started_at[MAX_CPUS];
+bool RESIZABLE_ARRAY(data, cpu_gimme_task);
+u64 RESIZABLE_ARRAY(data, cpu_started_at);
 
 struct central_timer {
 	struct bpf_timer timer;
@@ -189,7 +188,7 @@ void BPF_STRUCT_OPS(central_dispatch, s32 cpu, struct task_struct *prev)
 				break;
 
 			/* central's gimme is never set */
-			gimme = MEMBER_VPTR(cpu_gimme_task, [cpu]);
+			gimme = ARRAY_ELEM_PTR(cpu_gimme_task, cpu, nr_cpu_ids);
 			if (gimme && !*gimme)
 				continue;
 
@@ -220,7 +219,7 @@ void BPF_STRUCT_OPS(central_dispatch, s32 cpu, struct task_struct *prev)
 		if (scx_bpf_consume(FALLBACK_DSQ_ID))
 			return;
 
-		gimme = MEMBER_VPTR(cpu_gimme_task, [cpu]);
+		gimme = ARRAY_ELEM_PTR(cpu_gimme_task, cpu, nr_cpu_ids);
 		if (gimme)
 			*gimme = true;
 
@@ -235,7 +234,7 @@ void BPF_STRUCT_OPS(central_dispatch, s32 cpu, struct task_struct *prev)
 void BPF_STRUCT_OPS(central_running, struct task_struct *p)
 {
 	s32 cpu = scx_bpf_task_cpu(p);
-	u64 *started_at = MEMBER_VPTR(cpu_started_at, [cpu]);
+	u64 *started_at = ARRAY_ELEM_PTR(cpu_started_at, cpu, nr_cpu_ids);
 	if (started_at)
 		*started_at = bpf_ktime_get_ns() ?: 1;	/* 0 indicates idle */
 }
@@ -243,7 +242,7 @@ void BPF_STRUCT_OPS(central_running, struct task_struct *p)
 void BPF_STRUCT_OPS(central_stopping, struct task_struct *p, bool runnable)
 {
 	s32 cpu = scx_bpf_task_cpu(p);
-	u64 *started_at = MEMBER_VPTR(cpu_started_at, [cpu]);
+	u64 *started_at = ARRAY_ELEM_PTR(cpu_started_at, cpu, nr_cpu_ids);
 	if (started_at)
 		*started_at = 0;
 }
@@ -262,7 +261,7 @@ static int central_timerfn(void *map, int *key, struct bpf_timer *timer)
 			continue;
 
 		/* kick iff the current one exhausted its slice */
-		started_at = MEMBER_VPTR(cpu_started_at, [cpu]);
+		started_at = ARRAY_ELEM_PTR(cpu_started_at, cpu, nr_cpu_ids);
 		if (started_at && *started_at &&
 		    vtime_before(now, *started_at + SCX_SLICE_DFL))
 			continue;
--- a/tools/sched_ext/scx_central.c
+++ b/tools/sched_ext/scx_central.c
@@ -63,6 +63,10 @@ int main(int argc, char **argv)
 		}
 	}
 
+	/* Resize arrays so their element count is equal to cpu count. */
+	RESIZE_ARRAY(data, cpu_gimme_task, skel->rodata->nr_cpu_ids);
+	RESIZE_ARRAY(data, cpu_started_at, skel->rodata->nr_cpu_ids);
+
 	SCX_BUG_ON(scx_central__load(skel), "Failed to load skel");
 
 	link = bpf_map__attach_struct_ops(skel->maps.central_ops);
--- a/tools/sched_ext/scx_common.bpf.h
+++ b/tools/sched_ext/scx_common.bpf.h
@@ -82,6 +82,26 @@ SEC("struct_ops.s/"#name)							\
 BPF_PROG(name, ##args)
 
 /**
+ * RESIZABLE_ARRAY - Generates annotations for an array that may be resized
+ * @elfsec: the data section of the BPF program in which to place the array
+ * @arr: the name of the array
+ *
+ * libbpf has an API for setting map value sizes. Since data sections (i.e.
+ * bss, data, rodata) themselves are maps, a data section can be resized. If
+ * a data section has an array as its last element, the BTF info for that
+ * array will be adjusted so that length of the array is extended to meet the
+ * new length of the data section. This macro annotates an array to have an
+ * element count of one with the assumption that this array can be resized
+ * within the userspace program. It also annotates the section specifier so
+ * this array exists in a custom sub data section which can be resized
+ * independently.
+ *
+ * See RESIZE_ARRAY() for the userspace convenience macro for resizing an
+ * array declared with RESIZABLE_ARRAY().
+ */
+#define RESIZABLE_ARRAY(elfsec, arr) arr[1] SEC("."#elfsec"."#arr)
+
+/**
  * MEMBER_VPTR - Obtain the verified pointer to a struct or array member
  * @base: struct or array to index
  * @member: dereferenced member (e.g. ->field, [idx0][idx1], ...)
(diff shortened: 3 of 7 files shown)


tools/sched_ext/scx_common.bpf.h
```diff
@@ -115,6 +115,36 @@ BPF_PROG(name, ##args)
...
+ * Similar to MEMBER_VPTR() but accepts one additional size parameter which
+ * is used to compute the max allowable address within the given struct or
+ * array.
+ * It can be used in cases where a global array is defined with an initial
+ * size but is intended to be be resized before loading the bpf program.
+ * Without this version of the macro, MEMBER_VPTR() will use the compile time
+ * size of the array to compute the max, which will result in rejection by
+ * the verifier.
+ */
+#define MEMBER_VPTR_SZ(base, member, sz) (typeof(base member) *)({		\
```
anakryiko: this macro will only ever be used with arrays, most probably, so I wonder if it would be better to just have an array-specific macro ARRAY_ELEM_PTR(base, idx, cnt), where cnt is actual size of the array (so you can derive byte size as `sizeof(base[0]) * cnt`

tools/sched_ext/scx_example_central.bpf.c
```diff
@@ -189,7 +188,8 @@ void BPF_STRUCT_OPS(central_dispatch, s32 cpu, struct task_struct *prev)
 				break;
 
 			/* central's gimme is never set */
-			gimme = MEMBER_VPTR(cpu_gimme_task, [cpu]);
+			gimme = MEMBER_VPTR_SZ(cpu_gimme_task, [cpu],
```
anakryiko: and this could become just

```
gimme = ARRAY_ELEM_PTR(cpu_gimme_task, cpu, nr_cpu_ids);
```

tools/sched_ext/scx_example_central.c
```diff
@@ -64,6 +65,18 @@ int main(int argc, char **argv)
 		}
 	}
 
+	/* resize arrays so their element count is equal to cpu count.
+	 * after resizing, associated skel pointers are invalidated
+	 * so they need to be assigned the new value.
+	 */
+	assert(!bpf_map__set_value_size(skel->maps.data_cpu_gimme_task,
```
anakryiko: nit: assert() is probably an overkill here, I'd drop it in favor of more succinct code

tools/sched_ext/scx_example_pair.c
```diff
@@ -68,14 +69,35 @@ int main(int argc, char **argv)
 		}
 	}
 
+	bpf_map__set_max_entries(skel->maps.pair_ctx, skel->rodata->nr_cpu_ids);
```
anakryiko: original size was MAX_CPU/2, so should this be nr_cpu_ids/2 ?

tools/sched_ext/scx_example_pair.c
```diff
@@ -68,14 +69,35 @@ int main(int argc, char **argv)
...
-		if (skel->rodata->pair_cpu[i] < 0) {
-			skel->rodata->pair_cpu[i] = i + stride;
-			skel->rodata->pair_cpu[i + stride] = i;
-			skel->rodata->pair_id[i] = i;
-			skel->rodata->pair_id[i + stride] = i;
-			skel->rodata->in_pair_idx[i] = 0;
-			skel->rodata->in_pair_idx[i + stride] = 1;
+		if (skel->rodata_pair_cpu->pair_cpu[i] < 0) {
+			skel->rodata_pair_cpu->pair_cpu[i] = i + stride;
+			skel->rodata_pair_cpu->pair_cpu[i + stride] = i;
```
anakryiko: wouldn't `i+stride` overflow beyond array size? should it be `(i + stride) % nr_cpu_ids`?

tools/sched_ext/scx_common.bpf.h
inwardvessel: it's a good idea. changed to an array specific macro.

tools/sched_ext/scx_example_central.c
inwardvessel: fair enough. removed.

tools/sched_ext/scx_example_pair.c
inwardvessel: ok, I updated to be consistent. it originally seemed arbitrary since it was MAX_CPUS/2 (4096/2), so I didn't think it was necessary to divide to save space now that it was potentially much smaller (matching the actual cpu count). I also did consider a cpu count of 1 at the time but now I realize that this program needs at least 2 cpu's to function correctly. could be wrong, but I made an issue about this https://github.com/sched-ext/sched_ext/issues/28

tools/sched_ext/scx_example_pair.c
inwardvessel: I think this question is related more to the original code so I opened an issue for this #29 . not the definitive answer, but it seems that it does not overflow in the case where stride is `nr_cpu_ids / 2`, since the second half array elements become positive values before they are visited so the conditional is not entered for them. it seems in this case you could just iterate half the array and be done, but see the issue I linked because that might not be feasible for other stride values.

inwardvessel: resolved in `sched_ext` branch.

inwardvessel:
rebased on latest `sched_ext`

anakryiko (approved):
LGTM

inwardvessel:
rebased with no code changes

tools/sched_ext/scx_central.c
```diff
@@ -64,6 +65,18 @@ int main(int argc, char **argv)
 		}
 	}
 
+	/* resize arrays so their element count is equal to cpu count.
+	 * after resizing, associated skel pointers are invalidated
+	 * so they need to be assigned the new value.
```
ByteLab-David: Could you please make comments of the form:
```
/* 
 * Comment comment comment.
 * /
```

In other words, please match the existing style of the other schedulers, and please use correct grammar / punctuation, etc.

tools/sched_ext/scx_common.bpf.h
```diff
@@ -117,6 +117,34 @@ BPF_PROG(name, ##args)
...
+ *
+ * Similar to MEMBER_VPTR() but is intended for use with arrays where the
+ * element count needs to be explicit.
+ * It can be used in cases where a global array is defined with an initial
+ * size but is intended to be be resized before loading the bpf program.
+ * Without this version of the macro, MEMBER_VPTR() will use the compile time
+ * size of the array to compute the max, which will result in rejection by
+ * the verifier.
+ */
+#define ARRAY_ELEM_PTR(arr, i, n) (typeof(arr[i]) *)({		\
```
ByteLab-David: Could you please align the `\`'s?

tools/sched_ext/scx_pair.c
```diff
@@ -68,11 +69,32 @@ int main(int argc, char **argv)
 		}
 	}
 
+	bpf_map__set_max_entries(skel->maps.pair_ctx, skel->rodata->nr_cpu_ids / 2);
+
+	/* resize arrays so their element count is equal to cpu count.
```
ByteLab-David: Same here r.e. comment

Decave (commented):
So, I think it probably makes sense to land this for the sake of optimization and not having to hard-code MAX_CPUS. It does add some kind of odd-looking boilerplate, but I guess we'll have to live with it for now.

@htejun please chime in if you disagree or have other thoughts

Also, sorry @inwardvessel, but I think you may have to rebase one more time :-/

tools/sched_ext/scx_central.bpf.c
```diff
@@ -73,8 +72,8 @@ struct {
 } central_q SEC(".maps");
 
 /* can't use percpu map due to bad lookups */
-static bool cpu_gimme_task[MAX_CPUS];
-static u64 cpu_started_at[MAX_CPUS];
+bool cpu_gimme_task[1] SEC(".data.cpu_gimme_task");
+u64 cpu_started_at[1] SEC(".data.cpu_started_at");
```
ByteLab-David: Can we add a macro to hide the complexity here? Something like this:

```
#define RESIZABLE_ARRAY(__var) __var[1] SEC(".data.##__var")
```

tools/sched_ext/scx_central.c
```diff
@@ -64,6 +65,18 @@ int main(int argc, char **argv)
...
+	 */
+	bpf_map__set_value_size(skel->maps.data_cpu_gimme_task,
+				sizeof(skel->data_cpu_gimme_task->cpu_gimme_task[0]) * skel->rodata->nr_cpu_ids);
+	skel->data_cpu_gimme_task = bpf_map__initial_value(skel->maps.data_cpu_gimme_task, &sz);
+
+	bpf_map__set_value_size(skel->maps.data_cpu_started_at,
+				sizeof(skel->data_cpu_started_at->cpu_started_at[0]) * skel->rodata->nr_cpu_ids);
+	skel->data_cpu_started_at = bpf_map__initial_value(skel->maps.data_cpu_started_at, &sz);
```
ByteLab-David: Could you also please add a helper (with documentation) that encapsulates all of this reassignment boilerplate as well?

Decave (changes requested):

inwardvessel:
@Decave I made updates to address all comments. agreed on the boilerplate. at this point, the nature of resizing bpf arrays is really a side effect of resizing the actual data sections that contain them. so there's no direct way to say "resize this array". what we are really saying is "resize this map which happens to be the data section that contains the array I want resized". I added some macros based on your feedback. maybe they can be candidates for libbif itself so as to ease usage in cases like this.

tools/sched_ext/scx_user_common.h
```diff
@@ -31,4 +31,27 @@
...
+ * region.
+ */
+#define RESIZE_ARRAY(datasec, arr, n) \
+	do { \
+		size_t __sz; \
+		bpf_map__set_value_size(skel->maps.datasec##_##arr,  \
+				sizeof(skel->datasec##_##arr->arr[0]) * n); \
+		skel->datasec##_##arr = \
+			bpf_map__initial_value(skel->maps.datasec##_##arr, &__sz); \
+	} while (0)
```
ByteLab-David: nit: Could you please align the `\`'s?

tools/sched_ext/scx_pair.c
```diff
@@ -67,27 +67,33 @@ int main(int argc, char **argv)
...
+		if (skel->rodata_pair_cpu->pair_cpu[i] >= 0)
 			continue;
 
 		SCX_BUG_ON(i == j,
 			   "Invalid stride %d - CPU%d wants to be its own pair",
 			   stride, i);
 
-		SCX_BUG_ON(skel->rodata->pair_cpu[j] >= 0,
-			   "Invalid stride %d - three CPUs (%d, %d, %d) want to be a pair",
-			   stride, i, j, skel->rodata->pair_cpu[j]);
```
ByteLab-David: Did you mean to remove this check?

tools/sched_ext/scx_user_common.h
```diff
@@ -31,4 +31,27 @@
 			SCX_BUG((__fmt) __VA_OPT__(,) __VA_ARGS__);	\
 	} while (0)
 
+/**
+ * RESIZE_ARRAY - Convenience macro for resizing a bpf array
```
ByteLab-David: s/bpf/BPF

here and elsewhere in the comment

tools/sched_ext/scx_user_common.h
```diff
@@ -31,4 +31,27 @@
...
+ * @datasec: the data section of the bpf program in which to the array exists
```
ByteLab-David: s/data section/ELF section?

In general I wonder if this arg would be better named `elfsec`. What do you think?

tools/sched_ext/scx_user_common.h
```diff
@@ -31,4 +31,27 @@
...
+ * the array is adjusted so that the array length is sized to cover the new
+ * data section size. The second operation is reassigning the skeleton pointer
+ * for that custom data section so that it points to the newly memory mapped
+ * region.
+ */
+#define RESIZE_ARRAY(datasec, arr, n) \
+	do { \
+		size_t __sz; \
+		bpf_map__set_value_size(skel->maps.datasec##_##arr,  \
+				sizeof(skel->datasec##_##arr->arr[0]) * n); \
```
ByteLab-David: Could you please surround `n` in parentheses?

tools/sched_ext/scx_common.bpf.h
```diff
@@ -117,6 +137,34 @@ BPF_PROG(name, ##args)
...
+	u64 __base = (u64)arr;							\
+	u64 __addr = (u64)&(arr[i]) - __base;					\
+	asm volatile (								\
+		"if %0 <= %[max] goto +2\n"					\
+		"%0 = 0\n"							\
+		"goto +1\n"							\
+		"%0 += %1\n"							\
+		: "+r"(__addr)							\
+		: "r"(__base),							\
+		  [max]"r"(sizeof(arr[0]) * (n - 1)));				\
```
ByteLab-David: Could you please surround `n` in its own parens?

Decave (commented):
This is looking much better! Overall it's pretty close to landing, just had a few more nits / questions. Thanks for working on this.

tools/sched_ext/scx_pair.c
inwardvessel: good catch. reverted.

tools/sched_ext/scx_user_common.h
inwardvessel: sounds good, made the change.

Decave (approved):

PR #21, opened 2023-06-20
Bug fixes and a lot of atropos updates

htejun:
Fixes for a couple tricky bugs and an improvement to select_cpu_dfl().

*  All schedulers w/ vtime scheduling wasn't updating p->dsq.vtime on scheduler reloads or, if applicable, cgroup migrations. This could get some unlucky tasks stuck way in the future stalling them severely. It's likely that we haven't noticed this with web benchmarks because web machines never completely peg themselves.
* ttwu_queue path skips select_cpu() which can confuse some of the SCX schedulers. Fixed by disabling ttwu_queue for tasks on SCX.
* It turned out that going for a different CPU whose SMT siblings are all idle is usually better than sticking with partially free @prev_cpu. select_cpu_dfl() is updated accordingly.

Atropos saw a lot of updates:

* Execution stall bug caused by queueing on and waking different domains fixed.
* Load balancer updated so that it reliably and gradually reaches equilibrium.
* Tuner which runs every 100ms and tells the BPF portion to adapt to the current per-domain utilization level has been added. This allows turning on/off certain greedy execution optimizations based on CPU saturation state.
* PUSH/KICK_GREEDY optimizations implemented. Observed to improve work conservation and thus bandwidth w/ kcryptd benchmark. Would likely make latencies more consistent too.
* Improve handling of possible but not online CPUs.
* Other cleanups and improvements.

With these improvements, on an AMD zen2 CPU, atropos reliably outperforms CFS when running fio randrw workloads on a dm-crypted SSD across different concurrency levels.

### DIFF
--- a/kernel/sched/ext.c
+++ b/kernel/sched/ext.c
@@ -1933,8 +1933,7 @@ static s32 scx_pick_idle_cpu(const struct cpumask *cpus_allowed, u64 flags)
 	int cpu;
 
 retry:
-#ifdef CONFIG_SCHED_SMT
-	if (static_branch_likely(&sched_smt_present)) {
+	if (sched_smt_active()) {
 		cpu = cpumask_any_and_distribute(idle_masks.smt, cpus_allowed);
 		if (cpu < nr_cpu_ids) {
 			const struct cpumask *sbm = topology_sibling_cpumask(cpu);
@@ -1955,7 +1954,7 @@ retry:
 		if (flags & SCX_PICK_IDLE_CPU_WHOLE)
 			return -EBUSY;
 	}
-#endif
+
 	cpu = cpumask_any_and_distribute(idle_masks.cpu, cpus_allowed);
 	if (cpu >= nr_cpu_ids)
 		return -EBUSY;
@@ -1992,12 +1991,11 @@ static s32 scx_select_cpu_dfl(struct task_struct *p, s32 prev_cpu, u64 wake_flag
 	if (p->nr_cpus_allowed == 1)
 		return prev_cpu;
 
-#ifdef CONFIG_SCHED_SMT
 	/*
 	 * If CPU has SMT, any wholly idle CPU is likely a better pick than
 	 * partially idle @prev_cpu.
 	 */
-	if (static_branch_likely(&sched_smt_present)) {
+	if (sched_smt_active()) {
 		if (cpumask_test_cpu(prev_cpu, idle_masks.smt) &&
 		    test_and_clear_cpu_idle(prev_cpu)) {
 			p->scx.flags |= SCX_TASK_ENQ_LOCAL;
@@ -2010,7 +2008,6 @@ static s32 scx_select_cpu_dfl(struct task_struct *p, s32 prev_cpu, u64 wake_flag
 			return cpu;
 		}
 	}
-#endif
 
 	if (test_and_clear_cpu_idle(prev_cpu)) {
 		p->scx.flags |= SCX_TASK_ENQ_LOCAL;
--- a/kernel/sched/ext.c
+++ b/kernel/sched/ext.c
@@ -1932,7 +1932,9 @@ static s32 scx_pick_idle_cpu(const struct cpumask *cpus_allowed, u64 flags)
 {
 	int cpu;
 
-	do {
+retry:
+#ifdef CONFIG_SCHED_SMT
+	if (static_branch_likely(&sched_smt_present)) {
 		cpu = cpumask_any_and_distribute(idle_masks.smt, cpus_allowed);
 		if (cpu < nr_cpu_ids) {
 			const struct cpumask *sbm = topology_sibling_cpumask(cpu);
@@ -1947,17 +1949,22 @@ static s32 scx_pick_idle_cpu(const struct cpumask *cpus_allowed, u64 flags)
 				cpumask_andnot(idle_masks.smt, idle_masks.smt, sbm);
 			else
 				cpumask_andnot(idle_masks.smt, idle_masks.smt, cpumask_of(cpu));
-		} else {
-			if (flags & SCX_PICK_IDLE_CPU_WHOLE)
-				return -EBUSY;
-
-			cpu = cpumask_any_and_distribute(idle_masks.cpu, cpus_allowed);
-			if (cpu >= nr_cpu_ids)
-				return -EBUSY;
+			goto found;
 		}
-	} while (!test_and_clear_cpu_idle(cpu));
 
-	return cpu;
+		if (flags & SCX_PICK_IDLE_CPU_WHOLE)
+			return -EBUSY;
+	}
+#endif
+	cpu = cpumask_any_and_distribute(idle_masks.cpu, cpus_allowed);
+	if (cpu >= nr_cpu_ids)
+		return -EBUSY;
+
+found:
+	if (test_and_clear_cpu_idle(cpu))
+		return cpu;
+	else
+		goto retry;
 }
 
 static s32 scx_select_cpu_dfl(struct task_struct *p, s32 prev_cpu, u64 wake_flags)
@@ -1982,14 +1989,33 @@ static s32 scx_select_cpu_dfl(struct task_struct *p, s32 prev_cpu, u64 wake_flag
 		}
 	}
 
-	/* if the previous CPU is idle, dispatch directly to it */
+	if (p->nr_cpus_allowed == 1)
+		return prev_cpu;
+
+#ifdef CONFIG_SCHED_SMT
+	/*
+	 * If CPU has SMT, any wholly idle CPU is likely a better pick than
+	 * partially idle @prev_cpu.
+	 */
+	if (static_branch_likely(&sched_smt_present)) {
+		if (cpumask_test_cpu(prev_cpu, idle_masks.smt) &&
+		    test_and_clear_cpu_idle(prev_cpu)) {
+			p->scx.flags |= SCX_TASK_ENQ_LOCAL;
+			return prev_cpu;
+		}
+
+		cpu = scx_pick_idle_cpu(p->cpus_ptr, SCX_PICK_IDLE_CPU_WHOLE);
+		if (cpu >= 0) {
+			p->scx.flags |= SCX_TASK_ENQ_LOCAL;
+			return cpu;
+		}
+	}
+#endif
+
 	if (test_and_clear_cpu_idle(prev_cpu)) {
 		p->scx.flags |= SCX_TASK_ENQ_LOCAL;
 		return prev_cpu;
 	}
-
-	if (p->nr_cpus_allowed == 1)
-		return prev_cpu;
 
 	cpu = scx_pick_idle_cpu(p->cpus_ptr, 0);
 	if (cpu >= 0) {
--- a/kernel/sched/ext.c
+++ b/kernel/sched/ext.c
@@ -126,7 +126,6 @@ static struct {
 	cpumask_var_t smt;
 } idle_masks CL_ALIGNED_IF_ONSTACK;
 
-static bool __cacheline_aligned_in_smp scx_has_idle_cpus;
 #endif	/* CONFIG_SMP */
 
 /* for %SCX_KICK_WAIT */
@@ -1926,13 +1925,7 @@ void __scx_notify_pick_next_task(struct rq *rq, struct task_struct *task,
 
 static bool test_and_clear_cpu_idle(int cpu)
 {
-	if (cpumask_test_and_clear_cpu(cpu, idle_masks.cpu)) {
-		if (cpumask_empty(idle_masks.cpu))
-			scx_has_idle_cpus = false;
-		return true;
-	} else {
-		return false;
-	}
+	return cpumask_test_and_clear_cpu(cpu, idle_masks.cpu);
 }
 
 static s32 scx_pick_idle_cpu(const struct cpumask *cpus_allowed)
@@ -1978,7 +1971,7 @@ static s32 scx_select_cpu_dfl(struct task_struct *p, s32 prev_cpu, u64 wake_flag
 	 * local DSQ of the waker.
 	 */
 	if ((wake_flags & SCX_WAKE_SYNC) && p->nr_cpus_allowed > 1 &&
-	    scx_has_idle_cpus && !(current->flags & PF_EXITING)) {
+	    !cpumask_empty(idle_masks.cpu) && !(current->flags & PF_EXITING)) {
 		cpu = smp_processor_id();
 		if (cpumask_test_cpu(cpu, p->cpus_ptr)) {
 			p->scx.flags |= SCX_TASK_ENQ_LOCAL;
@@ -2045,7 +2038,6 @@ static void reset_idle_masks(void)
 	/* consider all cpus idle, should converge to the actual state quickly */
 	cpumask_setall(idle_masks.cpu);
 	cpumask_setall(idle_masks.smt);
-	scx_has_idle_cpus = true;
 }
 
 void __scx_update_idle(struct rq *rq, bool idle)
@@ -2061,8 +2053,6 @@ void __scx_update_idle(struct rq *rq, bool idle)
 
 	if (idle) {
 		cpumask_set_cpu(cpu, idle_masks.cpu);
-		if (!scx_has_idle_cpus)
-			scx_has_idle_cpus = true;
 
 		/*
 		 * idle_masks.smt handling is racy but that's fine as it's only
@@ -2075,9 +2065,6 @@ void __scx_update_idle(struct rq *rq, bool idle)
 		cpumask_or(idle_masks.smt, idle_masks.smt, sib_mask);
 	} else {
 		cpumask_clear_cpu(cpu, idle_masks.cpu);
-		if (scx_has_idle_cpus && cpumask_empty(idle_masks.cpu))
-			scx_has_idle_cpus = false;
-
 		cpumask_andnot(idle_masks.smt, idle_masks.smt, sib_mask);
 	}
 }
--- a/kernel/sched/ext.c
+++ b/kernel/sched/ext.c
@@ -4117,6 +4117,14 @@ bool scx_bpf_test_and_clear_cpu_idle(s32 cpu)
  * Pick and claim an idle cpu in @cpus_allowed. Returns the picked idle cpu
  * number on success. -%EBUSY if no matching cpu was found.
  *
+ * Idle CPU tracking may race against CPU scheduling state transitions. For
+ * example, this function may return -%EBUSY as CPUs are transitioning into the
+ * idle state. If the caller then assumes that there will be dispatch events on
+ * the CPUs as they were all busy, the scheduler may end up stalling with CPUs
+ * idling while there are pending tasks. Use scx_bpf_pick_any_cpu() and
+ * scx_bpf_kick_cpu() to guarantee that there will be at least one dispatch
+ * event in the near future.
+ *
  * Unavailable if ops.update_idle() is implemented and
  * %SCX_OPS_KEEP_BUILTIN_IDLE is not set.
  */
@@ -4128,6 +4136,37 @@ s32 scx_bpf_pick_idle_cpu(const struct cpumask *cpus_allowed, u64 flags)
 	}
 
 	return scx_pick_idle_cpu(cpus_allowed, flags);
+}
+
+/**
+ * scx_bpf_pick_any_cpu - Pick and claim an idle cpu if available or pick any CPU
+ * @cpus_allowed: Allowed cpumask
+ * @flags: %SCX_PICK_IDLE_CPU_* flags
+ *
+ * Pick and claim an idle cpu in @cpus_allowed. If none is available, pick any
+ * CPU in @cpus_allowed. Guaranteed to suceed and returns the picked idle cpu
+ * number if @cpus_allowed is not empty. -%EBUSY is returned if @cpus_allowed is
+ * empty.
+ *
+ * If ops.update_idle() is implemented and %SCX_OPS_KEEP_BUILTIN_IDLE is not
+ * set, this function can't tell which CPUs are idle and will always pick any
+ * CPU.
+ */
+s32 scx_bpf_pick_any_cpu(const struct cpumask *cpus_allowed, u64 flags)
+{
+	s32 cpu;
+
+	if (static_branch_likely(&scx_builtin_idle_enabled)) {
+		cpu = scx_pick_idle_cpu(cpus_allowed, flags);
+		if (cpu >= 0)
+			return cpu;
+	}
+
+	cpu = cpumask_any_distribute(cpus_allowed);
+	if (cpu < nr_cpu_ids)
+		return cpu;
+	else
+		return -EBUSY;
 }
 
 /**
@@ -4316,6 +4355,7 @@ BTF_ID_FLAGS(func, scx_bpf_kick_cpu)
 BTF_ID_FLAGS(func, scx_bpf_dsq_nr_queued)
 BTF_ID_FLAGS(func, scx_bpf_test_and_clear_cpu_idle)
 BTF_ID_FLAGS(func, scx_bpf_pick_idle_cpu, KF_RCU)
+BTF_ID_FLAGS(func, scx_bpf_pick_any_cpu, KF_RCU)
 BTF_ID_FLAGS(func, scx_bpf_get_idle_cpumask, KF_ACQUIRE)
 BTF_ID_FLAGS(func, scx_bpf_get_idle_smtmask, KF_ACQUIRE)
 BTF_ID_FLAGS(func, scx_bpf_put_idle_cpumask, KF_RELEASE)
--- a/kernel/sched/ext.c
+++ b/kernel/sched/ext.c
@@ -1925,7 +1925,25 @@ void __scx_notify_pick_next_task(struct rq *rq, struct task_struct *task,
 
 static bool test_and_clear_cpu_idle(int cpu)
 {
-	return cpumask_test_and_clear_cpu(cpu, idle_masks.cpu);
+	if (!cpumask_test_and_clear_cpu(cpu, idle_masks.cpu))
+		return false;
+
+	if (sched_smt_active()) {
+		const struct cpumask *sbm = topology_sibling_cpumask(cpu);
+
+		/*
+		 * If offline, @cpu is not its own sibling and
+		 * scx_pick_idle_cpu() can get caught in an infinite loop as
+		 * @cpu is never cleared from idle_masks.smt. Ensure that @cpu
+		 * is eventually cleared.
+		 */
+		if (cpumask_intersects(sbm, idle_masks.smt))
+			cpumask_andnot(idle_masks.smt, idle_masks.smt, sbm);
+		else if (cpumask_test_cpu(cpu, idle_masks.smt))
+			__cpumask_clear_cpu(cpu, idle_masks.smt);
+	}
+
+	return true;
 }
 
 static s32 scx_pick_idle_cpu(const struct cpumask *cpus_allowed, u64 flags)
@@ -1935,21 +1953,8 @@ static s32 scx_pick_idle_cpu(const struct cpumask *cpus_allowed, u64 flags)
 retry:
 	if (sched_smt_active()) {
 		cpu = cpumask_any_and_distribute(idle_masks.smt, cpus_allowed);
-		if (cpu < nr_cpu_ids) {
-			const struct cpumask *sbm = topology_sibling_cpumask(cpu);
-
-			/*
-			 * If offline, @cpu is not its own sibling and we can
-			 * get caught in an infinite loop as @cpu is never
-			 * cleared from idle_masks.smt. Clear @cpu directly in
-			 * such cases.
-			 */
-			if (likely(cpumask_test_cpu(cpu, sbm)))
-				cpumask_andnot(idle_masks.smt, idle_masks.smt, sbm);
-			else
-				cpumask_andnot(idle_masks.smt, idle_masks.smt, cpumask_of(cpu));
+		if (cpu < nr_cpu_ids)
 			goto found;
-		}
 
 		if (flags & SCX_PICK_IDLE_CPU_WHOLE)
 			return -EBUSY;
--- a/kernel/sched/ext.c
+++ b/kernel/sched/ext.c
@@ -1928,7 +1928,7 @@ static bool test_and_clear_cpu_idle(int cpu)
 	return cpumask_test_and_clear_cpu(cpu, idle_masks.cpu);
 }
 
-static s32 scx_pick_idle_cpu(const struct cpumask *cpus_allowed)
+static s32 scx_pick_idle_cpu(const struct cpumask *cpus_allowed, u64 flags)
 {
 	int cpu;
 
@@ -1948,6 +1948,9 @@ static s32 scx_pick_idle_cpu(const struct cpumask *cpus_allowed)
 			else
 				cpumask_andnot(idle_masks.smt, idle_masks.smt, cpumask_of(cpu));
 		} else {
+			if (flags & SCX_PICK_IDLE_CPU_WHOLE)
+				return -EBUSY;
+
 			cpu = cpumask_any_and_distribute(idle_masks.cpu, cpus_allowed);
 			if (cpu >= nr_cpu_ids)
 				return -EBUSY;
@@ -1988,7 +1991,7 @@ static s32 scx_select_cpu_dfl(struct task_struct *p, s32 prev_cpu, u64 wake_flag
 	if (p->nr_cpus_allowed == 1)
 		return prev_cpu;
 
-	cpu = scx_pick_idle_cpu(p->cpus_ptr);
+	cpu = scx_pick_idle_cpu(p->cpus_ptr, 0);
 	if (cpu >= 0) {
 		p->scx.flags |= SCX_TASK_ENQ_LOCAL;
 		return cpu;
@@ -2084,7 +2087,7 @@ static void rq_offline_scx(struct rq *rq, enum rq_onoff_reason reason)
 #else /* !CONFIG_SMP */
 
 static bool test_and_clear_cpu_idle(int cpu) { return false; }
-static s32 scx_pick_idle_cpu(const struct cpumask *cpus_allowed) { return -EBUSY; }
+static s32 scx_pick_idle_cpu(const struct cpumask *cpus_allowed, u64 flags) { return -EBUSY; }
 static void reset_idle_masks(void) {}
 
 #endif /* CONFIG_SMP */
@@ -3992,7 +3995,7 @@ static const struct btf_kfunc_id_set scx_kfunc_set_cpu_release = {
 /**
  * scx_bpf_kick_cpu - Trigger reschedule on a CPU
  * @cpu: cpu to kick
- * @flags: SCX_KICK_* flags
+ * @flags: %SCX_KICK_* flags
  *
  * Kick @cpu into rescheduling. This can be used to wake up an idle CPU or
  * trigger rescheduling on a busy CPU. This can be called from any online
@@ -4081,21 +4084,22 @@ bool scx_bpf_test_and_clear_cpu_idle(s32 cpu)
 /**
  * scx_bpf_pick_idle_cpu - Pick and claim an idle cpu
  * @cpus_allowed: Allowed cpumask
+ * @flags: %SCX_PICK_IDLE_CPU_* flags
  *
- * Pick and claim an idle cpu which is also in @cpus_allowed. Returns the picked
- * idle cpu number on success. -%EBUSY if no matching cpu was found.
+ * Pick and claim an idle cpu in @cpus_allowed. Returns the picked idle cpu
+ * number on success. -%EBUSY if no matching cpu was found.
  *
  * Unavailable if ops.update_idle() is implemented and
  * %SCX_OPS_KEEP_BUILTIN_IDLE is not set.
  */
-s32 scx_bpf_pick_idle_cpu(const struct cpumask *cpus_allowed)
+s32 scx_bpf_pick_idle_cpu(const struct cpumask *cpus_allowed, u64 flags)
 {
 	if (!static_branch_likely(&scx_builtin_idle_enabled)) {
 		scx_ops_error("built-in idle tracking is disabled");
 		return -EBUSY;
 	}
 
-	return scx_pick_idle_cpu(cpus_allowed);
+	return scx_pick_idle_cpu(cpus_allowed, flags);
 }
 
 /**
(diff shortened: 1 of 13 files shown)


kernel/sched/ext.c
```diff
@@ -1926,20 +1925,16 @@ void __scx_notify_pick_next_task(struct rq *rq, struct task_struct *task,
...
 
-static s32 scx_pick_idle_cpu(const struct cpumask *cpus_allowed)
+static s32 scx_pick_idle_cpu(const struct cpumask *cpus_allowed, u64 flags)
 {
 	int cpu;
 
-	do {
+retry:
+#ifdef CONFIG_SCHED_SMT
+	if (static_branch_likely(&sched_smt_present)) {
```
ByteLab-David: Can we use sched_smt_active() here and elsewhere that we're checking sched_smt_present directly?

kernel/sched/ext.c
```diff
@@ -1978,24 +1981,43 @@ static s32 scx_select_cpu_dfl(struct task_struct *p, s32 prev_cpu, u64 wake_flag
...
 		return prev_cpu;
+
+#ifdef CONFIG_SCHED_SMT
+	/*
+	 * If CPU has SMT, any wholly idle CPU is likely a better pick than
+	 * partially idle @prev_cpu.
+	 */
+	if (static_branch_likely(&sched_smt_present)) {
+		if (cpumask_test_cpu(prev_cpu, idle_masks.smt) &&
+		    test_and_clear_cpu_idle(prev_cpu)) {
```
ByteLab-David, line 2005: Shouldn't we do the same thing here with idle_masks.smt as we do in scx_pick_idle_cpu()?

kernel/sched/ext.c
```diff
@@ -2097,7 +2113,7 @@ static void rq_offline_scx(struct rq *rq, enum rq_onoff_reason reason)
 #else /* !CONFIG_SMP */
 
 static bool test_and_clear_cpu_idle(int cpu) { return false; }
-static s32 scx_pick_idle_cpu(const struct cpumask *cpus_allowed) { return -EBUSY; }
+static s32 scx_pick_idle_cpu(const struct cpumask *cpus_allowed, u64 flags) { return -EBUSY; }
```
ByteLab-David, line 2118: Should this be broken into multiple lines given its length?

```
static s32 scx_pick_idle_cpu(const struct cpumask *cpus_allowed, u64 flags)
{
        return -EBUSY;
}
```

kernel/sched/ext.h
```diff
@@ -91,6 +91,10 @@ enum scx_kick_flags {
 	SCX_KICK_WAIT		= 1LLU << 1,	/* wait for the CPU to be rescheduled */
 };
 
+enum scx_pick_idle_cpu_flags {
+	SCX_PICK_IDLE_CPU_WHOLE	= 1LLU << 0,	/* pick a CPU whose SMT siblings are also idle */
```
ByteLab-David, line 95: s/SMT siblings are/SMT sibling is (right?)

kernel/sched/ext.c
```diff
@@ -4094,21 +4110,22 @@ bool scx_bpf_test_and_clear_cpu_idle(s32 cpu)
...
  * Unavailable if ops.update_idle() is implemented and
  * %SCX_OPS_KEEP_BUILTIN_IDLE is not set.
  */
-s32 scx_bpf_pick_idle_cpu(const struct cpumask *cpus_allowed)
+s32 scx_bpf_pick_idle_cpu(const struct cpumask *cpus_allowed, u64 flags)
 {
 	if (!static_branch_likely(&scx_builtin_idle_enabled)) {
 		scx_ops_error("built-in idle tracking is disabled");
 		return -EBUSY;
 	}
```
ByteLab-David, line 4137: Should we also add a check here for invalid flags?

```
if (flags & ~SCX_PICK_IDLE_CPU_WHOLE) {
        scx_ops_error("Invalid pick idle flags %lx", flags);
        return -EINVAL;
}
```

tools/sched_ext/atropos/src/bpf/atropos.bpf.c
```diff
@@ -74,11 +74,14 @@ const volatile __u64 slice_ns = SCX_SLICE_DFL;
...
 
+/*
+ * Per-CPU context
+ */
 struct pcpu_ctx {
 	__u32 dom_rr_cur; /* used when scanning other doms */
 
 	/* libbpf-rs does not respect the alignment, so pad out the struct explicitly */
-	__u8 _padding[CACHELINE_SIZE - sizeof(u64)];
+	__u8 _padding[CACHELINE_SIZE - sizeof(u32)];
```
ByteLab-David, line 84: I thought Daniel said this was fixed in libbpf-rs 0.20.0?

tools/sched_ext/atropos/src/bpf/atropos.bpf.c
```diff
@@ -154,21 +227,24 @@ static bool task_set_dsq(struct task_ctx *task_ctx, struct task_struct *p,
...
-		scx_bpf_error("No dom%u", old_dom_id);
+		scx_bpf_error("Failed to lookup old dom%u", old_dom_id);
 		return false;
 	}
 
-	vtime_delta = p->scx.dsq_vtime - old_domc->vtime_now;
+	if (init_dsq_vtime)
+		vtime_delta = 0;
+	else
+		vtime_delta = p->scx.dsq_vtime - old_domc->vtime_now;
```
ByteLab-David, line 237: Hmm, could we just do this rather than adding a new field to the subprog?

```
if (p->scx.dsq_vtime)
        vtime_delta = p->scx.dsq_vtime - old_domc->vtime_now;
```

tools/sched_ext/atropos/src/bpf/atropos.bpf.c
```diff
@@ -194,22 +270,27 @@ static bool task_set_dsq(struct task_ctx *task_ctx, struct task_struct *p,
...
+	if (!(task_ctx = bpf_map_lookup_elem(&task_data, &pid)) ||
+	    !(p_cpumask = task_ctx->cpumask))
 		return -ENOENT;
 
 	if (kthreads_local &&
 	    (p->flags & PF_KTHREAD) && p->nr_cpus_allowed == 1) {
 		cpu = prev_cpu;
 		stat_add(ATROPOS_STAT_DIRECT_DISPATCH, 1);
-		goto local;
+		goto direct;
```
ByteLab-David, line 293: Does this successfully load? I'd have expected the verifier to reject it given that we call scx_bpf_put_idle_cpumask() on the `direct:` path without having previously called `scx_bpf_get_idle_smtmask()`

tools/sched_ext/atropos/src/bpf/atropos.bpf.c
```diff
@@ -234,7 +315,7 @@ s32 BPF_STRUCT_OPS(atropos_select_cpu, struct task_struct *p, int prev_cpu,
 			}
 			d_cpumask = domc->cpumask;
 			if (!d_cpumask) {
-				scx_bpf_error("Failed to acquire domain %u cpumask kptr",
+				scx_bpf_error("Failed to acquire dom%u cpumask kptr",
 					      task_ctx->dom_id);
 				return prev_cpu;
```
ByteLab-David: Doesn't really matter, but for the sake of consistency with e.g. the task_ctx lookup above, should we return -ENOENT anywhere that we don't find a kptr we expect?

tools/sched_ext/atropos/src/bpf/atropos.bpf.c
```diff
@@ -250,71 +331,185 @@ s32 BPF_STRUCT_OPS(atropos_select_cpu, struct task_struct *p, int prev_cpu,
...
+		if (direct_greedy_cpumask &&
+		    bpf_cpumask_test_cpu(prev_cpu, (const struct cpumask *)
+					 direct_greedy_cpumask) &&
+		    bpf_cpumask_test_cpu(prev_cpu, idle_smtmask) &&
+		    scx_bpf_test_and_clear_cpu_idle(prev_cpu)) {
+			stat_add(ATROPOS_STAT_GREEDY_IDLE, 1);
+			cpu = prev_cpu;
+			goto direct;
+		}
+	}
```
ByteLab-David: I'm having trouble understanding what the idea is with this path. Is it that if the task was previously greedily fetched by another domain with sufficiently low cpu util (and thus is being woken on a foreign domain), we should only keep it on the prev_cpu if that foreign domain is still sufficiently idle so as to support receiving remote tasks?

tools/sched_ext/atropos/src/bpf/atropos.bpf.c
```diff
@@ -250,71 +331,185 @@ s32 BPF_STRUCT_OPS(atropos_select_cpu, struct task_struct *p, int prev_cpu,
...
+
+	/*
+	 * @prev_cpu didn't work out. Find the best idle domestic CPU.
+	 */
+
+	/* If there is a domestic whole idle CPU, dispatch directly */
```
ByteLab-David, line 388: FWIW I think it may be useful to formally define domestic vs. foreign domains in the file's header comment. It can get a bit confusing when you consider that a task may be woken on a foreign domain. So "domestic" in this case is actually not the domain its local to at the time of wakeup. It all makes sense, it's just subtle and IMO could be easily confused.

tools/sched_ext/atropos/src/bpf/atropos.bpf.c
```diff
@@ -250,71 +331,185 @@ s32 BPF_STRUCT_OPS(atropos_select_cpu, struct task_struct *p, int prev_cpu,
...
+		cpu = scx_bpf_pick_idle_cpu((const struct cpumask *)p_cpumask,
+					    SCX_PICK_IDLE_CPU_WHOLE);
+		if (cpu >= 0) {
+			stat_add(ATROPOS_STAT_DIRECT_DISPATCH, 1);
+			goto direct;
+		}
+	}
+
+	/*
+	 * If @prev_cpu was domestic and is idle itself even though the whole
```
ByteLab-David, line 399: Can we add a comment explaining why this would only apply if @prev_cpu is domestic? I assume it's so that we quickly get back to equilibrium amongst domains?

tools/sched_ext/atropos/src/bpf/atropos.bpf.c
```diff
@@ -250,71 +331,185 @@ s32 BPF_STRUCT_OPS(atropos_select_cpu, struct task_struct *p, int prev_cpu,
...
+		}
+	}
+
+	/*
+	 * We're going to queue on the domestic domain's DSQ. @prev_cpu may be
+	 * in a different domain. Returning an out-of-domain CPU can lead to
+	 * stalls as all in-domain CPUs may be idle by the time @p gets
+	 * enqueued.
 	 */
 	if (bpf_cpumask_test_cpu(prev_cpu, (const struct cpumask *)p_cpumask))
```
ByteLab-David: We can just check prev_domestic here

tools/sched_ext/atropos/src/bpf/atropos.bpf.c
```diff
@@ -250,71 +331,185 @@ s32 BPF_STRUCT_OPS(atropos_select_cpu, struct task_struct *p, int prev_cpu,
...
 	new_dom = bpf_map_lookup_elem(&lb_data, &pid);
 	if (new_dom && *new_dom != task_ctx->dom_id &&
-	    task_set_dsq(task_ctx, p, *new_dom)) {
-		struct bpf_cpumask *p_cpumask;
-		s32 cpu;
-
+	    task_set_domain(task_ctx, p, *new_dom, false)) {
 		stat_add(ATROPOS_STAT_LOAD_BALANCE, 1);
 
 		/*
```
ByteLab-David: I don't get why this is necessary. Can't we just put the task directly on the local DSQ? Why would doing a direct dispatch like that cause it to stall? Or did you mean that CPUs in the newly-domestic domain could stall?

tools/sched_ext/atropos/src/bpf/atropos.bpf.c
```diff
@@ -327,15 +522,10 @@ void BPF_STRUCT_OPS(atropos_enqueue, struct task_struct *p, u32 enq_flags)
...
-			scx_bpf_error("Failed to get task_ctx->cpumask");
-			return;
-		}
-		cpu = scx_bpf_pick_idle_cpu((const struct cpumask *)p_cpumask);
-
+		cpu = scx_bpf_pick_idle_cpu((const struct cpumask *)p_cpumask, 0);
 		if (cpu >= 0)
 			scx_bpf_kick_cpu(cpu, 0);
+		goto dom_queue;
 	}
```
ByteLab-David: Is this necessary to avoid stalls? Or is it an SIS_NODE-like optimization where we're checking at enqueue time whether there's another core available in the current domain, and then kicking it so we can dispatch to it on the balance path?

tools/sched_ext/atropos/src/bpf/atropos.bpf.c
```diff
@@ -344,6 +534,23 @@ void BPF_STRUCT_OPS(atropos_enqueue, struct task_struct *p, u32 enq_flags)
 		return;
 	}
 
+	/*
+	 * @p is about to be queued on its domain's dsq. However, @p may be on a
+	 * foreign CPU due to a greedy execution and not have gone through
+	 * ->select_cpu() if it's being enqueued e.g. after slice exhaustion. If
+	 * so, @p would be queued on its domain's dsq but none of the CPUs in
+	 * the domain would be woken up for it which can induce execution
+	 * bubles. Kick a domestic CPU if @p is on a foreign domain.
```
ByteLab-David: bubbles? not sure what that means exactly, perhaps a different word would be better?

also, yeah, this makes sense. nice debugging.

tools/sched_ext/atropos/src/bpf/atropos.bpf.c
```diff
@@ -344,6 +534,23 @@ void BPF_STRUCT_OPS(atropos_enqueue, struct task_struct *p, u32 enq_flags)
...
+	 */
+	if (!bpf_cpumask_test_cpu(scx_bpf_task_cpu(p), (const struct cpumask *)p_cpumask)) {
+		cpu = scx_bpf_pick_idle_cpu((const struct cpumask *)p_cpumask, 0);
+		if (cpu < 0)
+			cpu = bpf_cpumask_any((const struct cpumask *)p_cpumask);
+		scx_bpf_kick_cpu(cpu, 0);
```
ByteLab-David: Why do we need to always kick a CPU, even if no idle CPU is found?

tools/sched_ext/atropos/src/bpf/atropos.bpf.c
```diff
@@ -442,15 +659,16 @@ void BPF_STRUCT_OPS(atropos_dispatch, s32 cpu, struct task_struct *prev)
...
 
-	if (!task_ctx) {
-		scx_bpf_error("No task_ctx[%d]", pid);
+	if (!(task_ctx = bpf_map_lookup_elem(&task_data, &pid))) {
+		scx_bpf_error("Failed to lookup task_ctx");
 		return;
 	}
 
 	task_ctx->runnable_at = bpf_ktime_get_ns();
+	task_ctx->is_kworker = p->flags & PF_WQ_WORKER;
```
ByteLab-David, line 668: Any reason this can't go in .prep_enable()?

tools/sched_ext/atropos/src/bpf/atropos.bpf.c
```diff
@@ -511,86 +729,70 @@ void BPF_STRUCT_OPS(atropos_quiescent, struct task_struct *p, u64 deq_flags)
...
+		dom = (dom + 1) % nr_doms;
+		if (cpumask_intersects_domain(cpumask, dom)) {
+			task_ctx->dom_mask |= 1LLU << dom;
+			if (first_dom == MAX_DOMS)
+				first_dom = dom;
+		}
+	}
 
-	return lctx.dom_id;
+	return first_dom;
```
ByteLab-David, line 765: Should we add a comment explaining that while we're always returning the first domain, that the load balancer should eventually bring equilibrium across all domains?

tools/sched_ext/atropos/src/bpf/atropos.bpf.c
```diff
@@ -696,24 +898,60 @@ static int create_dom_dsq(u32 idx, void *data)
...
+	s32 i, ret;
+
 	if (!switch_partial)
 		scx_bpf_switch_all();
 
-	bpf_loop(nr_doms, create_dom_dsq, NULL, 0);
+	bpf_for(i, 0, nr_doms) {
+		ret = create_dom(i);
+		if (ret)
+			return ret;
```
ByteLab-David, line 936: Feel free to ignore, but FWIW I really like the if (var = ...) syntax as you've used elsewhere:

```
if (ret = create_dom(i))
         return ret;
```

Decave (commented):
Awesome job with all the debugging. Overall the changes look great, just left some comments on a first read-through. Haven't looked at main.rs yet

kernel/sched/ext.h
htejun: They're all pairs now but it's possible to have multiple logical CPUs per physical core. IIRC, sun niagara was like that.

kernel/sched/ext.c
htejun: I kinda prefer keeping these blanks on a single line because they don't carry any information but yeah breaking it up into multiple lines is fine too.

kernel/sched/ext.c
htejun: Oh yeah, good point. Lemme fix that up.

htejun: So, this turned out to impact other stuff too - e.g. atropos' select_cpu. I moved the smt handling into test_and_clear_cpu_idle().

kernel/sched/ext.c
htejun: I'm not sure we want to preemptively sanitize input parameters. CPU number and enq_flags have values that can harm the kernel, so we explicitly protect against them but here there's nothing to protect yet, and I'm not sure testing against currently defined flags provides much value especially given that they would in most cases be read from BTF and will get undefined error that way.

tools/sched_ext/atropos/src/bpf/atropos.bpf.c
htejun: It didn't build. Some dep problem. Didn't wanna debug it now.

tools/sched_ext/atropos/src/bpf/atropos.bpf.c
htejun: Oh yeah, that looks like a verifier bug.

tools/sched_ext/atropos/src/bpf/atropos.bpf.c
htejun: Yeah, pushing tasks to busy remote domains lowers performance because those CPUs are gonna become busy pretty soon anyway. Pushing tasks doesn't really push up utilization but just increases noise (I push, others push to me, round and round).

tools/sched_ext/atropos/src/bpf/atropos.bpf.c
htejun: Yeah, we used to use "local" for this too but because we use "local" for CPUs, it got really confusing. We'll add a comment.

tools/sched_ext/atropos/src/bpf/atropos.bpf.c
htejun: It's just that there's inherent cost to going foreign and we're several lungs down in the optimization steps, so it now becomes difficult to measure whether something is better or worse.

tools/sched_ext/atropos/src/bpf/atropos.bpf.c
htejun: We might be okay without it but the problem is that the idle state transitions aren't interlocked with actual rq state transitions unlike kicking.

Let's say the target domain has only one CPU and the CPU is in the process of going idle. It hasn't cleared the idle bit yet but is just about to. pick_idle_cpu() will miss it, so we don't kick nobody and the CPU can then set idle and go to sleep while there's task pending. Kicking however is properly interlocked - ie. it's guaranteed that the CPU is going to look for tasks to run at least once after the last kicking.

We can make idle masks strongly interlocked by adding a scan between idle assertion and actually sleeping but let's get to that some other time.

tools/sched_ext/atropos/src/bpf/atropos.bpf.c
htejun: It gets set afterwards, unfortunately.

tools/sched_ext/atropos/src/bpf/atropos.bpf.c
htejun: We're round-robining the starting point, so this should spread out by default.

tools/sched_ext/atropos/src/bpf/atropos.bpf.c
htejun: So, kernel coding style guideline is against that. For BPF tho, because we have so many more, often redundant, condition checks required by the verifier, I thought it'd make sense to use the patern for checking conditions that we know to not happen.

I think the reason the style guide advises against it is because there's reasonable chance that one might write `if (A = B)` while intending to write `if (A == B)`. To add visual distinction, compiler warns on `if (A = B)` but is okay with `if ((A = B ))`, which is usually the used form.

So, yeah, I think we can probably establish a pattern which is a bit different from kernel and use the combined if expression more.

kernel/sched/ext.c
htejun: I'm just gonna keep it one line.

tools/sched_ext/atropos/src/bpf/atropos.bpf.c
htejun: But @p->scx.dsq_vtime may be any value. So, we can always set them to zero from core code side and then depend on it; however, that closes the possiblity for reloading a scheduler and then keeping using the existing states so that impact on scheduling is minimized. I'm not entirely sure whether that's something worthwhile to consider tho. Deciding to re-init them to 0 on load is at least an option we can opt for later.

kernel/sched/ext.c
ByteLab-David: So, this is something that's commonly done in BPF helpers that take flags. Doesn't mean we have to do it here, but there is a precedent for it in other BPF infra. Regarding BTF -- I may be missing something, but I don't think the verifier would do any sanitation of flags here, no? Wouldn't it just ensure that a scalar is being passed?

kernel/sched/ext.c
ByteLab-David: Sounds good to me

tools/sched_ext/atropos/src/bpf/atropos.bpf.c
ByteLab-David: Heh, and to your point, I lazily did the wrong version. Anyways, that all sounds good

tools/sched_ext/atropos/src/bpf/atropos.bpf.c
htejun: Okay fixed.

tools/sched_ext/atropos/src/bpf/atropos.bpf.c
htejun: Yeah, let's do that.

tools/sched_ext/atropos/src/bpf/atropos.bpf.c
ByteLab-David: Ah ok, the reloading point makes sense. Agreed that we can just leave it like this for now

tools/sched_ext/atropos/src/bpf/atropos.bpf.c
ByteLab-David: Ah I see. Yeah that's pretty subtle. IMO would be good to clean that up at some point, but not important for now.

tools/sched_ext/atropos/src/bpf/atropos.bpf.c
ByteLab-David: Ah, sorry, totally missed `dom_rr_cur`.

tools/sched_ext/atropos/src/main.rs
```diff
@@ -8,40 +8,57 @@ pub use atropos::*;
...
 /// balance the domains.
-
+///
 /// This scheduler demonstrates dividing scheduling logic between BPF and
 /// userspace and using rust to build the userspace part. An earlier variant of
 /// this scheduler was used to balance across six domains, each representing a
 /// chiplet in a six-chiplet AMD processor, and could match the performance of
 /// production setup using CFS.
+///
+/// WARNING: Atropos currenlty assumes that all domains have equal
```
ByteLab-David: s/currenlty/currently

tools/sched_ext/atropos/src/bpf/atropos.bpf.c
htejun: Added more comments.

tools/sched_ext/atropos/src/bpf/atropos.bpf.c
htejun: Updated to "temporary execution stalls".

Decave (commented):
Just 1 very minor nit for main.rs. Honestly kinda skimmed over it, but it looks reasonable

tools/sched_ext/atropos/src/bpf/atropos.bpf.c
htejun: Added a comment.

kernel/sched/ext.c
htejun: Oh I mean, if we're worried about users passing in future flags into older kernels, which would be the most common case where this check would trigger, it should choke on BTF. Yeah, I don't know. Maybe we'll decide to santize all inputs but right now it's more on the "check what's needed to protect the kernel" side, I think. Let's revisit later.

ByteLab-David: Yep, fine with leaving it off. I still don't really see how BTF would prevent it from being loaded, but in any case it's totally safe to pass any random value, so I agree it doesn't really protect anything and we don't need to validate anything.

tools/sched_ext/atropos/src/bpf/atropos.bpf.c
htejun: Picking an idle CPU in the new domain and kicking is necessary because otherwise all of their CPUs can sit idle while @p is sitting in its DSQ.

htejun: Although, as before, it should always kick something, even if nothing seems idle at the moment. We likely want to add scx_bpf_pick_a_cpu() which always returns a CPU.

tools/sched_ext/atropos/src/bpf/atropos.bpf.c
htejun: I'm not sure about this one. We don't want to directly dispatch to the local DSQ in general to honor the migration. It can be argued that similar logic as DIRECT_GREEDY can be used here but even if there's gain from that that's gonna be really tiny given how much lower frequency LB is compared to regular scheduling paths, so keeping it simple makes sense.

So, ignoring direct dispatch makes sense, but I'm not quite sure why we need to kick the CPU to avoid stalls. It's woken up and is going to check local and domain DSQ's and then go idle again reasserting its bit on the idle mask. All that seems like they should work fine without external kicking. It was added by the following commit in the old tree:

```
commit e856a0bc572bbfb6d680fbddf5f6a19f71bf2ae8
Author: Tejun Heo <tj@kernel.org>
Date:   Tue Mar 7 23:25:48 2023 -1000

    atropos: Kick cpu when ignoring dispatch_local while enqueuing

diff --git a/tools/sched_ext/atropos/src/bpf/atropos.bpf.c b/tools/sched_ext/atropos/src/bpf/atropos.bpf.c
index 10c916d227f2..3a2aeca137fe 100644
--- a/tools/sched_ext/atropos/src/bpf/atropos.bpf.c
+++ b/tools/sched_ext/atropos/src/bpf/atropos.bpf.c
@@ -289,7 +289,16 @@ void BPF_STRUCT_OPS(atropos_enqueue, struct task_struct *p, u32 enq_flags)
                s32 cpu;
 
                stat_add(ATROPOS_STAT_LOAD_BALANCE, 1);
-               task_ctx->dispatch_local = false;
+
+               /*
+                * If dispatch_local is set, We own @p's idle state but we are
+                * not gonna put the task in its local dsq which can cause the
+                * CPU to stall. Kick it.
+                */
+               if (task_ctx->dispatch_local) {
+                       task_ctx->dispatch_local = false;
+                       scx_bpf_kick_cpu(scx_bpf_task_cpu(p), 0);
+               }
 
                p_cpumask = bpf_cpumask_kptr_get(&task_ctx->cpumask);
                if (!p_cpumask) {
```

It could just be that I noticed some stalling and was confused about why they were happening. Let's remove it later. It doesn't hurt anything and I don't want to risk re-introducing stalls right now. I'll mark it with a comment.

tools/sched_ext/atropos/src/main.rs
```diff
@@ -894,6 +1119,40 @@ impl<'a> Scheduler<'a> {
...
+                next_sched_at += self.sched_interval;
+                if next_sched_at < now {
+                    next_sched_at = now + self.sched_interval;
+                }
+            }
+
+            std::thread::sleep(
+                next_sched_at
+                    .min(next_tune_at)
+                    .duration_since(std::time::SystemTime::now())?,
```
ByteLab-David, line 1150: @htejun if I run rcutorture with atropos, the scheduler gets kicked out after a minute or so with "Error: second time provided was later than self" from, presumably, here. It looks like SystemTime.duration_since() isn't really safe to use according to https://doc.rust-lang.org/std/time/struct.SystemTime.html#method.duration_since

>This function may fail because measurements taken earlier are not guaranteed to always be before later measurements (due to anomalies such as the system clock being adjusted either forwards or backwards). [Instant](https://doc.rust-lang.org/std/time/struct.Instant.html) can be used to measure elapsed time without this risk of failure.

Should we rework this to use Instant and Duration? Or to make it resilient to clock skew?

kernel/sched/ext.c
```diff
@@ -1926,42 +1925,50 @@ void __scx_notify_pick_next_task(struct rq *rq, struct task_struct *task,
...
 		return false;
+
+	if (sched_smt_active()) {
+		const struct cpumask *sbm = topology_sibling_cpumask(cpu);
+
+		/*
+		 * If offline, @cpu is not its own sibling and
+		 * scx_pick_idle_cpu() can get caught in an infinite loop as
+		 * @cpu is never cleared from idle_masks.smt. Ensure that @cpu
+		 * is eventually cleared.
```
dschatzberg, line 1938: Can you help me understand this change a bit more? What does this test and clear have to do with smt siblings?

kernel/sched/ext.c
```diff
@@ -4094,21 +4112,61 @@ bool scx_bpf_test_and_clear_cpu_idle(s32 cpu)
...
+	return scx_pick_idle_cpu(cpus_allowed, flags);
+}
+
+/**
+ * scx_bpf_pick_any_cpu - Pick and claim an idle cpu if available or pick any CPU
+ * @cpus_allowed: Allowed cpumask
+ * @flags: %SCX_PICK_IDLE_CPU_* flags
+ *
+ * Pick and claim an idle cpu in @cpus_allowed. If none is available, pick any
+ * CPU in @cpus_allowed. Guaranteed to suceed and returns the picked idle cpu
```
dschatzberg, line 4147: nit: suceed => succeed

tools/sched_ext/atropos/src/bpf/atropos.bpf.c
```diff
@@ -139,13 +143,82 @@ struct {
...
+ * that can be used directly in the scheduling paths.
+ */
+struct tune_input{
+	__u64 gen;
+	__u64 direct_greedy_cpumask[MAX_CPUS / 64];
+	__u64 kick_greedy_cpumask[MAX_CPUS / 64];
+} tune_input;
+
+__u64 tune_params_gen;
+private(A) struct bpf_cpumask __kptr *direct_greedy_cpumask;
```
dschatzberg, line 158: I'm not familiar with this private(A) syntax - what does it do?

tools/sched_ext/atropos/src/bpf/atropos.bpf.c
```diff
@@ -250,92 +331,200 @@ s32 BPF_STRUCT_OPS(atropos_select_cpu, struct task_struct *p, int prev_cpu,
...
 			}
 		}
 	}
 
-	/* if the previous CPU is idle, dispatch directly to it */
-	if (scx_bpf_test_and_clear_cpu_idle(prev_cpu)) {
-		stat_add(ATROPOS_STAT_PREV_IDLE, 1);
-		cpu = prev_cpu;
-		goto local;
-	}
```
dschatzberg, line 264: Just confirming - you say that removing this prev-idle check improved perf?

dschatzberg (commented):
Haven't finished looking through main.rs - but went through the bpf logic and it makes sense - few comments inline

tools/sched_ext/atropos/src/main.rs
vnepogodin: I think use of `Instant` seems more reasonable here. As code don't actually require "full" time to check if interval elapsed.

ByteLab-David: @vnepogodin yep, that was merged in https://github.com/sched-ext/sched_ext/commit/52db350b405153d1bc5cbc619060d4fc360b230e

kernel/sched/ext.c
htejun: So, there are two levels of idle masks, cpumaks and smtmask. cpumask is the real one and tracks the idle cpus. smtmask is the hint mask that tracks whether the whole core is idle. Let's say the toplogy is [0, 1], [2, 3] where the bracketed pairs are SMT siblings.

* If all are idle, both cpumask and smtmask would be 1111.
* If then CPU0 becomes busy, cpumask=0111 smtmask=0011
* If then CPU1 becomes busy, cpumask=0011 smtmask=0011
* If then CPU0 becomes idle, cpumask=1011 smtmask=0011
* If then CPU1 becomes idle, cpumask=1111 smtmask=1111

ie. the bits for the smtmask get cleared for the whole physical core when any of its logical threads are busy. Now, this is implemented by clearing what the kernel thinks SMT mask is - `topology_sibling_cpumask()` - whenever a CPU becomes busy and asserting the same mask on idle transition if all siblings are idle.

Now, select_cpu does sth like the following:

```
retry:
    if (there's a whole core idle) {
        clear the sibling group in the smtmask;
        try test_and_set that CPU in cpumask, if fails, goto retry;
    }
    if (there's an idle cpu)
        try test_and_set that CPU in cpumask, if fails, goto retry;
```

The problem is that `topology_sibling_cpumask()` of a CPU becomes empty when a CPU is offline. So, the "clearing the sibling group" part doesn't do anything and the the test_and_set part may fail because smtmask can go out of sync with cpumask, which puts us in an infinite loop. So, we clear `cpumask_of(cpu)` instead in that case to guarantee forward progress.

tools/sched_ext/atropos/src/bpf/atropos.bpf.c
htejun: It hides it from the automatic global mmapped access from userland. Otherwise, bpf gets unhappy with putting kptrs (allocated cpumask) there because userspace can clobber the value.

tools/sched_ext/atropos/src/bpf/atropos.bpf.c
htejun: So, yeah, that was my initial observation. For whatever reason, removing that noticeably improved performance. However, what the new code does now is to keep the foreign prev_idle if the whole smt pair is idle and that actually gains performance.

tools/sched_ext/atropos/src/main.rs
```diff
@@ -123,6 +149,328 @@ fn clear_map(map: &mut libbpf_rs::Map) {
...
+        .take((nr_cpus + 64) / 64)
+        .rev()
+        .fold(String::new(), |acc, x| format!("{} {:016X}", acc, x))
+}
+
+// Neither procfs or fb_procfs can determine per-CPU utilization reliably
+// with CPU hot[un]plugs. Roll our own.
+//
+// https://github.com/eminence/procfs/issues/274
+// https://github.com/facebookincubator/below/issues/8190
```
dschatzberg, line 164: @brianc118 fixed this in below this week. I'll move us back onto fb_procfs unless you have any concerns.

tools/sched_ext/atropos/src/main.rs
```diff
@@ -123,6 +149,328 @@ fn clear_map(map: &mut libbpf_rs::Map) {
...
+        let mut cpu_dom = vec![];
+
+        for cpu in 0..nr_cpus {
+            match cpu_to_cache[cpu] {
+                Some(cache_id) => {
+                    let dom_id = cache_to_dom[&cache_id];
+                    dom_cpus[dom_id].set(cpu, true);
+                    cpu_dom.push(Some(dom_id));
+                }
+                None => {
```
dschatzberg, line 377: In what case can you lack a cpu -> cache mapping?

tools/sched_ext/atropos/src/main.rs
htejun: Yeah, that's fine but let's do that next week after sending out the patchset.

tools/sched_ext/atropos/src/main.rs
htejun: So, on my zen2 test machine:
```
root@rbw /s/d/s/cpu# cat /proc/cpuinfo | grep name | head -1
model name      : AMD Ryzen 9 3900X 12-Core Processor
root@rbw ~# cat /sys/devices/system/cpu/possible
0-31
root@rbw ~# cat /sys/devices/system/cpu/present
0-23
root@rbw ~# cat /sys/devices/system/cpu/online
0-23
root@rbw ~# cd /sys/devices/system/cpu
root@rbw /s/d/s/cpu# cat possible present online
0-31
0-23
0-23
root@rbw /s/d/s/cpu# ls -d cpu*/cache | wc -l
24
```

The CCXs have some CPUs disabled. They can't be brought online but the firmware still reports them as possible but not present as if they can become present. So, we end up with possible CPUs without cache information. The same thing happens with some CCXs on belgamos IIRC. No idea why firmware is reporting it like that. It's silly because we end up allocating for the impossible possible CPUs and iterating over them too.

htejun: But I mean, in general, if hotplug is possible, then, the cache information may not be known until the CPU is plugged in. Even CPUs which are plugged in but brought down by echoing 0 to online don't report cache information. So, we need to be able to handle them. Eventually, I think what we should do is just reloading the scheduler when hotplugged CPUs are detected.

ptr1337: > cat /sys/devices/system/cpu/possible

@Decave 

Seems to be the same issue as I have reported to you with my 5900X.

@htejun 
I remember that this "issue" was not present in 6.2 Kernels, but in 6.1 and 6.3/6.4.

But good there is a solution for atropos! I give in the next time a testing again. Thanks for working on it!

PR #27, opened 2023-06-23
Misc updates

htejun:
Misc changes from v4 patchset prepping:

* Dropped _example prefix from scheduler names. It wasn't clear what _example was indicating and all it was doing was making the names longer.
* SCHED_CHANGE_BLOCK dropped as upstream is in the process of adopting something more generic.
* Some other cosmetic changes.

No functional changes intended.


kernel/sched/core.c
```diff
@@ -12121,3 +12156,38 @@ void sched_mm_cid_fork(struct task_struct *t)
...
+void sched_deq_and_put_task(struct task_struct *p, int queue_flags,
+			    struct sched_enq_and_set_ctx *ctx)
+{
+	struct rq *rq = task_rq(p);
+
+	lockdep_assert_rq_held(rq);
+
+	*ctx = (struct sched_enq_and_set_ctx){
+		.p = p,
+		.queue_flags = queue_flags | DEQUEUE_NOCLOCK,
```
ByteLab-David: Doesn't this extra DEQUEUE_NOLOCK cause a behavior change for the sched_deq_and_put_task() in scx_ops_disable_workfn()? Before we added ENQUEUE_NOCLOCK to the flags in sched_change_guard_fini().

Decave (approved):

kernel/sched/core.c
htejun: Ah shit, yeah, that's a mistake. Lemme drop that.

kernel/sched/ext.c
```diff
@@ -2984,10 +2993,11 @@ static void scx_ops_disable_workfn(struct kthread_work *work)
...
 		if (READ_ONCE(p->__state) != TASK_DEAD) {
-			SCHED_CHANGE_BLOCK(task_rq(p), p,
-					   DEQUEUE_SAVE | DEQUEUE_MOVE) {
-				/* cycling deq/enq is enough, see above */
-			}
+			struct sched_enq_and_set_ctx ctx;
+
+			/* cycling deq/enq is enough, see above */
+			sched_deq_and_put_task(p, DEQUEUE_SAVE | DEQUEUE_MOVE, &ctx);
+			sched_enq_and_set_task(&ctx);
```
ByteLab-David, line 3000: I think this is still a change in behavior -- we always set `ENQUEUE_NOCLOCK` on the sched_change_guard_fini() path before, so if we want to keep that we need to add it to sched_enq_and_set_task().

htejun: Indeed. Thanks for pointing that out.

PR #33, opened 2023-07-14
scx: atropos: Use fb-procfs crate

dschatzberg:
fb-procfs crate supports hotplug CPUs now, so migrate back to that for measuring CPU util.


htejun:
I merged it but it's kinda sad that the diffstat is +123-77 and the code overall looks more complicated. Maybe this indicates that fb_procfs can improve its interface?

PR #35, opened 2023-07-19
scx: Use explicit checks instead of asserts

Decave:
On some build systems, asserts are compiled out. We're currently using asserts on some actual operations, such as loading a prog. Let's update all of our progs to explicitly check return values where appropriate so we don't accidentally compile out real logic.

### DIFF
--- a/tools/sched_ext/scx_user_common.h
+++ b/tools/sched_ext/scx_user_common.h
@@ -1,0 +1,34 @@
+/* SPDX-License-Identifier: GPL-2.0 */
+/*
+ * Copyright (c) 2023 Meta Platforms, Inc. and affiliates.
+ * Copyright (c) 2023 Tejun Heo <tj@kernel.org>
+ * Copyright (c) 2023 David Vernet <dvernet@meta.com>
+ */
+#ifndef __SCHED_EXT_USER_COMMON_H
+#define __SCHED_EXT_USER_COMMON_H
+
+#include <stdarg.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+#ifdef __KERNEL__
+#error "Should not be included by BPF programs"
+#endif
+
+#define SCX_BUG(__fmt, ...)							\
+	do {									\
+		fprintf(stderr, "%s:%d [scx panic]: %s\n", __FILE__, __LINE__,	\
+			strerror(errno));					\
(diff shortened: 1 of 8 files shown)


htejun:
I'm not a fan of the changes. These error conditions shouldn't really happen and the explicit error handling code just adds clutters. What build system compiles out asserts? Can't we just add build flags / macros for those environments?

ByteLab-David:
This is BUCK. asserts are compiled out when NDEBUG is defined at the time of compilation, as described in https://www.tutorialspoint.com/c_standard_library/assert_h.htm. I don't think this is uncommon, and disabling this behavior in BUCK is probably a non-starter. We could always #undef NDEBUG, but that seems unnecessary / hacky.

>These error conditions shouldn't really happen and the explicit error handling code just adds clutters

Well, in this case, we're not really checking true invariants, with the one exception possibly being bpf_map__fd(). I'm fine with us erroring out on any of these conditions rather than printing a message and returning, but we should manually add some kind of panic() function if we want to do that instead of using assert.

htejun:
Can't we add a macro somewhere? It doesn't have to be named assert but the ratio of boilerplate is already rather high and I don't want to increase them.

ByteLab-David:
@htejun let me know if this looks better

tools/sched_ext/scx_user_common.h
```diff
@@ -0,0 +1,32 @@
...
+			strerror(errno));					\
+		fprintf(stderr, __fmt __VA_OPT__(,) __VA_ARGS__);		\
+		fprintf(stderr, "\n");						\
+										\
+		exit(EXIT_FAILURE);						\
+	} while (0)
+
+#define SCX_BUG_ON(__cond, __fmt, ...)	\
+	if (__cond)			\
+		SCX_BUG((__fmt) __VA_OPT__(,) __VA_ARGS__)
```
htejun: This probably should be wrapped inside do{}while(0) otherwise, something like the following can become confusing:
```
if (cond)
  SCX_BUG_ON(other_cond, ...);
else
  do something else;
```

htejun (approved):
Minor suggestion on the macro definition. Looks good to me otherwise.

tools/sched_ext/scx_user_common.h
ByteLab-David: Yep, sorry for the silly oversight.

PR #36, opened 2023-07-20
scx: Fix some clippy warnings in Atropos

Decave:
If you run clippy on atropos, it notices a few places where can clean up the rust code. Let's do that.


tools/sched_ext/scx_atropos/src/main.rs
```diff
@@ -659,18 +659,13 @@ impl<'a, 'b, 'c> LoadBalancer<'a, 'b, 'c> {
...
-            })
-            .next()
-        {
-            Some((OrderedFloat(load), task)) => Some((*load, task)),
-            None => None,
-        }
+            .find(|(_, task)| {
+                !task.migrated.get() &&
+                ((task.dom_mask & 1 << pull_dom) != 0) &&
+                (!skip_kworkers || !task.is_kworker)
```
htejun: So, I get that this is identity conversion but the logic is harder to understand this way. Most of the trivial converions look fine but I'm not sure we want to strictly follow clippy for higher level conversions.

ByteLab-David: Sounds good, I'll revert this change and leave the others

ByteLab-David: Fixed

htejun (approved):

PR #37, opened 2023-07-20
scx: Rename scx_atropos to scx_rusty

Decave:
scx_rusy is (kind of) a more descriptive name of the scheduler, so let's rename it to that.


ByteLab-David:
@htejun updated to scx_rusty

htejun (approved):

PR #39, opened 2023-08-01
scx: Adjust a couple of small things in rusty.bpf.c

Decave:
rusty.bpf.c has a few small places where we can improve either the formatting of the code, or the logic. In rusty_select_cpu(), we declare the idle_smtmask as struct cpumask *, when it could be const. Also, when initializing the pcpu_ctx, we're using an actual for-loop instead of bpf_for. Let's just fix up these small issues.

### DIFF
--- a/tools/sched_ext/scx_rusty/src/bpf/rusty.bpf.c
+++ b/tools/sched_ext/scx_rusty/src/bpf/rusty.bpf.c
@@ -273,7 +273,7 @@ static bool task_set_domain(struct task_ctx *task_ctx, struct task_struct *p,
 s32 BPF_STRUCT_OPS(rusty_select_cpu, struct task_struct *p, s32 prev_cpu,
 		   u64 wake_flags)
 {
-	struct cpumask *idle_smtmask = scx_bpf_get_idle_smtmask();
+	const struct cpumask *idle_smtmask = scx_bpf_get_idle_smtmask();
 	struct task_ctx *task_ctx;
 	struct bpf_cpumask *p_cpumask;
 	pid_t pid = p->pid;
@@ -933,7 +933,7 @@ s32 BPF_STRUCT_OPS_SLEEPABLE(rusty_init)
 			return ret;
 	}
 
-	for (u32 i = 0; i < nr_cpus; i++)
+	bpf_for(i, 0, nr_cpus)
 		pcpu_ctx[i].dom_rr_cur = i;
 
 	cpumask = bpf_cpumask_create();


tools/sched_ext/scx_rusty/src/bpf/rusty.bpf.c
```diff
@@ -933,8 +933,7 @@ s32 BPF_STRUCT_OPS_SLEEPABLE(rusty_init)
 			return ret;
 	}
 
-	for (u32 i = 0; i < nr_cpus; i++)
-		pcpu_ctx[i].dom_rr_cur = i;
+	bpf_for(i, 0, nr_cpus) pcpu_ctx[i].dom_rr_cur = i;
```
ByteLab-David: Note that this is on one line because clang-format complains if we split it into two

htejun: Is this better tho? It just looks like clang-format doesn't understand what's going on.

ByteLab-David: I agree it looks bad. I didn't love this change, which is why I highlighted it. We can drop it, and just agree to ignore certain clang-format requests that don't make sense when it's being built in an environment where clang-format is run (+cc @dschatzberg as an FYI)

htejun: Let's say this scx lands upstream and someone sends a patch. I don't think we can't tell that person to change the formatting because clang-format or other tools we used internally don't fully agree with what they're doing. If we want to sync the internal copy to the upstream one, the internal copy will just have to accept some deviations coming from upstream.

ByteLab-David:
@htejun fixed

tools/sched_ext/scx_rusty/src/bpf/rusty.bpf.c
dschatzberg: FWIW - the clang-format configuration we use internally is the exact same as the one already in the Linux tree (at the root there is a .clang-format).

I get that in Linux folks typically don't enforce that clang-format has been run. We have the option of opting-out internally as well but on the whole I find the cost of the occasional odd formatting to be worth having automation do the work.

I think accepting deviation here is fine

ByteLab-David: >I get that in Linux folks typically don't enforce that clang-format has been run. We have the option of opting-out internally as well but on the whole I find the cost of the occasional odd formatting to be worth having automation do the work.

Yeah, I don't think I've ever seen a patch review where someone requested that clang-format be run. Unfortunately, AFAIK the only thing people typically enforce is that check_patch.pl is run, which is pretty sad.

htejun: So, in kernel, people tend not to mind too much about not running or ignoring recommendations from tools but get bothered a lot if something made through automation is silly even when they are far and few between. ie. There's a very high cost attached to doing something obviously or intuitively silly in few places which often tilts the decision process away from applying any automated tooling in a uniform way.

There are pros and cons to both sides of argument but this isn't a fight we can win or even want to have.

htejun: Put differently, just like some repos enforce "formatting must follow this tooling", kernel enforces "it shouldn't look stupid" just as strongly, which is less clearly defined but is still enforced with the same vigor.

dschatzberg: Yeah, I understand the kernel stance and I'm not suggesting we should fight this. I'm saying that I would prefer the opposite for our internal codebase and accept that in this manner our internal and kernel codebases will diverge.

PR #48, opened 2023-09-21
scx_rusty: keep .bpf.o files for debugging

htejun:


anakryiko:
seems like you merged way more than just keeping .bpf.o, I don't know if this was intentional or not.

htejun:
Hahaha, not intentional. Lemme fix it up.

Issue #56, opened 2023-10-11
Various build issues


daandemeyer:
Grab bag of issues I discovered while integrating this into [mkosi-kernel](https://github.com/DaanDeMeyer/mkosi-kernel):

1. Build artifacts are not stored under the out-of-tree build directory if one is used (`O=`, `KBUILD_OUTPUT=`).
2. `cargo` is invoked as part of the build and will download dependencies, meaning the build needs network access. It should be possible to do all things that require network access up-front in a separate step
3. The examples cannot be installed, this means the kernel source tree with the built examples in it has to be mounted into the VM. It'd be great if the examples could be installed using `make install` instead.

ByteLab-David:
Thanks for the report, I'll work on adding support for these today

ByteLab-David:
@DaanDeMeyer this should all be addressed now. Could you please take a look and let me know if you need any other changes? If not, please feel free to close the issue.

daandemeyer:
`cargo` is still trying to update the index when I run it after downloading the dependencies. I think there's no way to check if the "download dependencies" cargo step has been executed so make will always try to run it?

ByteLab-David:
Hmmm, ok. We could always add a special build-specific make target that doesn't take the `cargo fetch` target step as a dependency. Would this do the trick?

```
diff --git a/tools/sched_ext/Makefile b/tools/sched_ext/Makefile
index fc2dca246d72..df93d3e098c0 100644
--- a/tools/sched_ext/Makefile
+++ b/tools/sched_ext/Makefile
@@ -203,9 +203,17 @@ scx_rusty_deps: $(SCX_COMMON_DEPS)
 scx_rusty: export RUSTFLAGS = -C link-args=-lzstd -C link-args=-lz -C link-args=-lelf -L $(BPFOBJ_DIR)
 scx_rusty: export SCX_RUSTY_CLANG = $(CLANG)
 scx_rusty: export SCX_RUSTY_BPF_CFLAGS = $(BPF_CFLAGS)
+
+define ccrusty
+ cargo build --manifest-path=$1/Cargo.toml --offline $(CARGOFLAGS)
+ $(Q)cp $(OUTPUT_DIR)/release/$1 $(BINDIR)/$1
+endef
+
 scx_rusty: $(INCLUDE_DIR)/vmlinux.h scx_rusty_deps
-   cargo build --manifest-path=$@/Cargo.toml --offline $(CARGOFLAGS)
[...]

PR #62, opened 2023-10-16
rusty: Further tweak build system

Decave:
We previously separated the scx_rusty build into two steps -- a step to download dependencies, and another to build. That mostly works, except that the download-dependencies step is always run before the build step as it's a dependency. Even though it doesn't download any cargo dependencies, it still accesses the network.

Let's add a way for builders to pass --offline to cargo via a CARGO_OFFLINE make variable.


ByteLab-David:
Addresses https://github.com/sched-ext/sched_ext/issues/56

daandemeyer:
I'll take this for a spin

daandemeyer:
This does the trick, thanks for fixing this!

PR #65, opened 2023-10-31
sched_ext: Add scx_layered

htejun:
See `scx_layered --help` for details.


tools/sched_ext/Makefile
```diff
@@ -218,12 +218,23 @@ scx_rusty: $(INCLUDE_DIR)/vmlinux.h $(SCX_COMMON_DEPS)
...
 
+scx_layered_deps:
+	cargo fetch --manifest-path=scx_layered/Cargo.toml
+
+scx_layered: export RUSTFLAGS = -C link-args=-lzstd -C link-args=-lz -C link-args=-lelf -L $(BPFOBJ_DIR)
+scx_layered: export SCX_LAYERED_CLANG = $(CLANG)
+scx_layered: export SCX_LAYERED_BPF_CFLAGS = $(BPF_CFLAGS)
+scx_layered: $(INCLUDE_DIR)/vmlinux.h $(SCX_COMMON_DEPS)
+	cargo build --manifest-path=$@/Cargo.toml $(CARGOFLAGS)
+	$(Q)cp $(OUTPUT_DIR)/release/$@ $(BINDIR)/$@
```
ByteLab-David: Can we write a Makefile function here instead of copy-pasting the cargo build recipe?

tools/sched_ext/scx_layered/src/bpf/layered.bpf.c
```diff
@@ -0,0 +1,947 @@
...
+private(all_cpumask) struct bpf_cpumask __kptr *all_cpumask;
+struct layer layers[MAX_LAYERS];
+u32 fallback_cpu;
+u32 preempt_cursor;
+
+#define dbg(fmt, args...)	do { if (debug) bpf_printk(fmt, ##args); } while (0)
+#define trace(fmt, args...)	do { if (debug > 1) bpf_printk(fmt, ##args); } while (0)
+
+#include "util.bpf.c"
+#include "ravg.bpf.c"
```
ByteLab-David: No need to block or waste time on this now, but at some point we should update the build system to use bpftool to link separate .bpf.o object files. That will make it less brittle than having to include the raw .bpf.c file. Although, I guess then we'd run into the issue of not being able to inline stuff.

tools/sched_ext/scx_layered/src/bpf/layered.bpf.c
```diff
@@ -0,0 +1,947 @@
...
+#include <bpf/bpf_tracing.h>
+
+char _license[] SEC("license") = "GPL";
+
+const volatile u32 debug = 0;
+const volatile u64 slice_ns = SCX_SLICE_DFL;
+const volatile u32 nr_possible_cpus = 1;
+const volatile u32 nr_layers = 1;
+const volatile bool smt_enabled = true;
+const volatile unsigned char all_cpus[MAX_CPUS_U8];
```
ByteLab-David, line 19: Can you use the RESIZABLE_ARRAY stuff here instead of hard-coding the max? Or, meh, I guess we only have convenience macros for that stuff for C schedulers :-(

tools/sched_ext/scx_layered/src/bpf/layered.bpf.c
```diff
@@ -0,0 +1,947 @@
...
+			bpf_cpumask_release(cpumask);
+	}
+
+	return 0;
+}
+
+void BPF_STRUCT_OPS(layered_exit, struct scx_exit_info *ei)
+{
+	uei_record(&uei, ei);
+}
```
ByteLab-David, line 955: It would be nice if we could somehow enable this as default behavior. I believe every single scheduler does it, and there's no reason not to want it. But yeah, don't really see how we could.

tools/sched_ext/scx_layered/src/bpf/layered.bpf.c
```diff
@@ -0,0 +1,947 @@
...
+			if (!cont || !cont->cpumask) {
+				scx_bpf_error("can't happen");
+				return;
+			}
+			if (*u8_ptr & (1 << (cpu % 8))) {
+				bpf_cpumask_set_cpu(cpu, cont->cpumask);
+				total++;
+			} else {
+				bpf_cpumask_clear_cpu(cpu, cont->cpumask);
+			}
```
ByteLab-David: Might be a bit faster to do a `bpf_cpumask_clear()` once before the loop rather than clearing each cpu individually?

tools/sched_ext/scx_layered/src/bpf/layered.bpf.c
```diff
@@ -0,0 +1,947 @@
...
+}
+
+SEC("fentry/scheduler_tick")
+int scheduler_tick_fentry(const void *ctx)
+{
+	u64 now;
+	int idx;
+
+	if (bpf_get_smp_processor_id() != 0)
+		return 0;
```
ByteLab-David: It seems kinda sad to always penalize CPU 0, but yeah, probably simpler and less error prone than trying to toss it around.

tools/sched_ext/scx_layered/src/bpf/layered.bpf.c
```diff
@@ -0,0 +1,947 @@
...
+
+	now = bpf_ktime_get_ns();
+	bpf_for(idx, 0, nr_layers) {
+		layers[idx].load_avg = ravg_read(&layer_loads[idx].ravg_data,
+						 now, USAGE_HALF_LIFE);
```
ByteLab-David: Might be a dumb question, but is there a reason we couldn't just use a BPF timer here? Why do we have to use fentry on scheduler_tick()? This is something we'd want to do even if ticks are disabled, no?

tools/sched_ext/scx_layered/src/bpf/layered.bpf.c
```diff
@@ -0,0 +1,947 @@
...
+			tctx->refresh_layer = true;
+	}
+
+	if (next)
+		bpf_task_release(next);
+	return 0;
+}
+
+SEC("fentry/__set_task_comm")
+int BPF_PROG(fentry_set_task_comm, struct task_struct *p, const char *buf, bool exec)
```
ByteLab-David: Can we instead use the task_rename trace point?

tools/sched_ext/scx_layered/src/bpf/layered.bpf.c
```diff
@@ -0,0 +1,947 @@
...
+		tctx->refresh_layer = true;
+}
+
+void BPF_STRUCT_OPS(layered_set_cpumask, struct task_struct *p,
+		    const struct cpumask *cpumask)
+{
+	struct task_ctx *tctx;
+	pid_t pid = p->pid;
+
+	if ((tctx = bpf_map_lookup_elem(&task_ctxs, &pid)) && all_cpumask)
```
ByteLab-David: lookup_task_ctx_may_fail()?

tools/sched_ext/scx_layered/src/bpf/layered.bpf.c
```diff
@@ -0,0 +1,947 @@
...
+
+		if (!(cont = bpf_map_lookup_elem(&layer_cpumasks, &i)))
+			return -ENONET;
+
+		cpumask = bpf_cpumask_create();
+		if (!cpumask)
+			return -ENOMEM;
+		cpumask = bpf_kptr_xchg(&cont->cpumask, cpumask);
+		if (cpumask)
+			bpf_cpumask_release(cpumask);
```
ByteLab-David, line 946: Can we call refresh_cpumasks() here? Not following why we'd want to wait for the first scheduler tick. I guess it's because we haven't called update_bpf_layer_cpumask() yet in rust? If so, maybe we should do that as part of the init process before we attach?

tools/sched_ext/scx_layered/src/bpf/layered.h
```diff
@@ -0,0 +1,96 @@
...
+enum consts {
+	MAX_CPUS_SHIFT		= 9,
+	MAX_CPUS		= 1 << MAX_CPUS_SHIFT,
+	MAX_CPUS_U8		= MAX_CPUS / 8,
+	MAX_TASKS		= 131072,
+	MAX_PATH		= 4096,
+	MAX_COMM		= 16,
+	MAX_LAYER_MATCH_ORS	= 32,
+	MAX_LAYERS		= 16,
+	USAGE_HALF_LIFE		= 1 * 100000000,	/* 100ms */
```
ByteLab-David: Can we drop the `1 *`?

tools/sched_ext/scx_layered/src/bpf/layered.bpf.c
```diff
@@ -0,0 +1,947 @@
...
+const volatile u64 slice_ns = SCX_SLICE_DFL;
+const volatile u32 nr_possible_cpus = 1;
+const volatile u32 nr_layers = 1;
+const volatile bool smt_enabled = true;
+const volatile unsigned char all_cpus[MAX_CPUS_U8];
+
+private(all_cpumask) struct bpf_cpumask __kptr *all_cpumask;
+struct layer layers[MAX_LAYERS];
+u32 fallback_cpu;
+u32 preempt_cursor;
```
ByteLab-David: static?

tools/sched_ext/scx_layered/src/bpf/layered.bpf.c
```diff
@@ -0,0 +1,947 @@
...
+		u32 cpu = (preempt_cursor + idx) % nr_possible_cpus;
+
+		if (!all_cpumask ||
+		    !bpf_cpumask_test_cpu(cpu, (const struct cpumask *)all_cpumask))
+			continue;
+		if (!(cand_cctx = lookup_cpu_ctx(cpu)) || cand_cctx->current_preempt)
+			continue;
+
+		scx_bpf_kick_cpu(cpu, SCX_KICK_PREEMPT);
+		preempt_cursor = (cpu + 1) % nr_possible_cpus;
```
ByteLab-David, line 491: Can you add a comment explaining that it's OK for these accesses to be racy / that this is just a rough heuristic?

tools/sched_ext/scx_layered/src/bpf/layered.bpf.c
```diff
@@ -0,0 +1,947 @@
...
+		if (!(layer_cpumask = lookup_layer_cpumask(idx)))
+			return;
+
+		if (bpf_cpumask_test_cpu(cpu, layer_cpumask)) {
+			if (scx_bpf_consume(idx))
+				return;
+		} else if (cpu == fallback_cpu && layer->nr_cpus == 0) {
+			if (scx_bpf_consume(idx))
+				return;
+		}
```
ByteLab-David: Can this perhaps be written more succinctly as follows:
```
if (bpf_cpumask_test_cpu(cpu, layer_cpumask) ||
    (cpu == fallback_cpu && layer->nr_cpus == 0)) {
			if (scx_bpf_consume(idx))
				return;
}
```

tools/sched_ext/scx_layered/src/bpf/layered.bpf.c
```diff
@@ -0,0 +1,947 @@
...
+}
+
+void BPF_STRUCT_OPS(layered_dispatch, s32 cpu, struct task_struct *prev)
+{
+	int idx;
+
+	/* consume preempting layers first */
+	bpf_for(idx, 0, nr_layers)
+		if (layers[idx].preempt && scx_bpf_consume(idx))
+			return;
```
ByteLab-David, line 505: Do you think it would make a difference to consume !open preempting layers first? Just wondering because we consume !preempting open layers last below.

tools/sched_ext/scx_layered/src/bpf/ravg.bpf.c
```diff
@@ -0,0 +1,329 @@
+/* to be included in the main bpf.c file */
+
+#define RAVG_FN_ATTRS		inline __attribute__((unused, always_inline))
+//#define RAVG_FN_ATTRS		__attribute__((unused))
```
ByteLab-David: Stray comment?

tools/sched_ext/scx_layered/src/main.rs
```diff
@@ -0,0 +1,1635 @@
...
+///     ]
+///   ],
+///
+/// The outer array contains the OR groups and the inner AND blocks, so the
+/// above matches:
+///
+/// * Tasks which are in the cgroup sub-hierarchy under "system.slice".
+/// * Or tasks whose comm starts with "fbagent" and have a nice value > 0.
+///
+/// Currenlty, the following matches are supported:
```
ByteLab-David: Currently

tools/sched_ext/scx_layered/src/main.rs
```diff
@@ -0,0 +1,1635 @@
...
+///
+/// * CgroupPrefix: Matches the prefix of the cgroup that the task belongs
+///   to. As this is a string match, whether the pattern has the trailing
+///   '/' makes difference. For example, "TOP/CHILD/" only matches tasks
```
ByteLab-David: s/makes difference/makes a difference

tools/sched_ext/scx_layered/src/main.rs
```diff
@@ -0,0 +1,1635 @@
...
+/// * CommPrefix: Matches the task's comm prefix.
+///
+/// * NiceAbove: Matches if the task's nice value is greater than the
+///   pattern.
+///
+/// * NiceBelow: Matches if the task's nice value is smaller than the
+///   pattern.
+///
+/// While there are complexity limitations as the matches are performed in
+/// BPF, it is straight-forward to add more types of matches.
```
ByteLab-David: s/straight-forward/straightforward

tools/sched_ext/scx_layered/src/main.rs
```diff
@@ -0,0 +1,1635 @@
...
+/// Policies
+/// ========
+///
+/// The following is an example policy configuration for a layer.
+///
+///   "kind": {
+///     "Confined": {
+///       "cpus_range": [1, 8],
+///       "util_range": [0.8, 0.9],
+///       ]
```
ByteLab-David: Let's remove the stray `]` to make this valid JSON

tools/sched_ext/scx_layered/src/main.rs
```diff
@@ -0,0 +1,1635 @@
...
+///   layers when no idle CPUs are available.
+///
+/// * Open: Prefer the CPUs which are not occupied by Confined or Grouped
+///   layers. Tasks in this group will spill into occupied CPUs if there are
+///   no unoccupied idle CPUs. If "preempt" is true, tasks in this layer
+///   will preempt tasks which belong to other non-preempting layers when no
+///   idle CPUs are available.
+///
+/// Similar to matches, adding new policies and extending existing ones
+/// should be relatively straight-forward.
```
ByteLab-David: straightforward

tools/sched_ext/scx_layered/src/main.rs
```diff
@@ -0,0 +1,1635 @@
...
+/// Note that the last layer in the configuration must have an empty match
+/// set as it must match all tasks which haven't been matched into previous
+/// layers.
+///
+/// The configuration can be specified in multiple json files and command
+/// line arguments. Each must contain valid layer configurations and they're
+/// concatenated in the specified order. In most cases, something like the
+/// following should do.
+///
+///   $ scx_layered file:example.json
```
ByteLab-David, line 172: Can we give an example of a full configuration which includes both matches and policies?

tools/sched_ext/scx_layered/src/main.rs
```diff
@@ -0,0 +1,1635 @@
...
+    ///
+    /// The paths don't have the leading '/' and may or may not have trailing
+    /// '/'. If there is no trailing '/', the prefix matches any cgroups
+    /// which have matching prefix upto that point.
+    ///
+    /// - "" matches all cgroups.
+    /// - "/" only matches the root cgroup.
+    /// - "workload" matches both "workload/work" and "workload-1/work".
+    /// - "workload/" matches "workload/work" but not "workload-1/work".
+    specs: Vec<String>,
```
ByteLab-David: Can we also add a description for how to specify a layer's policies?

tools/sched_ext/scx_layered/src/bpf/layered.h
```diff
@@ -0,0 +1,96 @@
...
+	int		nice_above_or_below;
+};
+
+struct layer_match_ands {
+	struct layer_match	matches[NR_LAYER_MATCH_KINDS];
+	int			nr_match_ands;
+};
+
+struct layer {
+	struct layer_match_ands	matches[MAX_LAYER_MATCH_ORS];
```
ByteLab-David, line 82: Apologies for the bikeshedding, but can we avoid using the same `matches` field name for both of these types? This feels especially confusing because a `struct layer_match_ands` object is really an OR instance from the perspective of `struct layer`. At first I wasn't a huge fan of using ORs and ANDs to describe these objects in general because they seem like qualifiers describing the same type of object, but after reading your description writeup for the command I think it's intuitive. From a code perspective though, this is a bit confusing for readability.

Maybe even just something like the following?

```

struct layer_match_ands {
	struct layer_match	ands[NR_LAYER_MATCH_KINDS];
	int			nr_match_ands;
};

struct layer {
	struct layer_match_ands	ors[MAX_LAYER_MATCH_ORS];
```

Decave (commented):
Awesome job on this. Looks mostly good, just left a bunch of comments / nits.

tools/sched_ext/scx_layered/src/bpf/layered.bpf.c
ByteLab-David: Meant to reference `u32 preempt_cursor` specifically

tools/sched_ext/scx_layered/src/bpf/layered.bpf.c
```diff
@@ -0,0 +1,947 @@
...
+
+	bpf_for(cpu, 0, nr_possible_cpus) {
+		u8 *u8_ptr;
+
+		if ((u8_ptr = MEMBER_VPTR(layers, [idx].cpus[cpu / 8]))) {
+			barrier_var(cont);
+			if (!cont || !cont->cpumask) {
+				scx_bpf_error("can't happen");
+				return;
+			}
```
ByteLab-David: Can we check this once outside of the loop?

tools/sched_ext/Makefile
ByteLab-David: This should be taken care of if you rebase onto https://github.com/sched-ext/sched_ext/pull/67

tools/sched_ext/scx_layered/src/bpf/layered.bpf.c
htejun: Yeah, right now, the usability isn't all that great. Unlike, e.g., max tasks, these are relatively small and not really a limiting factor, so maybe later.

tools/sched_ext/scx_layered/src/bpf/layered.bpf.c
htejun: We need to call these under spinlock, so bpf side isn't ready yet. But yeah eventually we need these to be proper libraries.

tools/sched_ext/scx_layered/src/bpf/layered.bpf.c
htejun: verifier gets unhappy if I place that anywhere else. I'll mark it and ping bpf folks.

tools/sched_ext/scx_layered/src/bpf/layered.bpf.c
htejun: Maybe, but this isn't a hot path and memory ordering is kinda iffy. On the writer side, we're okay because we're concluding with fetch_and_add; however, on the reader side, we're just doing naked read without any ordering constraint. We can add full mb'd operation there but that's a hot path, so it's a bit excessive. So, keeping all ops atomic means that we at least ensure that only bits which really need to change are flipped. Hmm... the mask update is cold path, so we can add a full mb there after the initial check, but idk, maybe just leave it as-is w/ some comments and update later when we get proper load/store ops?

htejun: Oh I mean, load_acquire/store_release.

htejun: Ah, and we're testing which bit is set without locking in select_cpu(). We actually need to flip only the bits which need to flipped. If we pre-clear them, we'll trigger AFFN_VIOL path unnecessarily.

tools/sched_ext/scx_layered/src/bpf/layered.bpf.c
htejun: CPU 0 is always online. Other CPUs may go down, so CPU 0 is already special and treated differently in some places.

tools/sched_ext/scx_layered/src/bpf/layered.bpf.c
htejun: Yeah, timer should be fine too. Just wanted to try using ticks. Now that I wrote ravg_read is rust, I probably don't need periodic update anyway. Will update.

tools/sched_ext/scx_layered/src/bpf/layered.bpf.c
htejun: Will update.

tools/sched_ext/scx_layered/src/bpf/layered.bpf.c
htejun: Preempting implies !confined. Whether distinguishing open vs. grouped would be meaningful, idk. Right now, there's no ordering between preempting layers. We can get to that question when we need them?

tools/sched_ext/scx_layered/src/bpf/layered.bpf.c
htejun: Applied.

tools/sched_ext/scx_layered/src/bpf/layered.bpf.c
htejun: Hopefully, we will come up with a way to crate'ize helpers in the future.

tools/sched_ext/scx_layered/src/bpf/ravg.bpf.c
htejun: Imma make it use the shared ravg headers.

tools/sched_ext/scx_layered/src/main.rs
htejun: I thought about that but the help message is pretty long already and I'm not sure including the whole file inline adds all that much when the example file can easily be generated.

tools/sched_ext/scx_layered/src/main.rs
htejun: This is an earlier comment which should have been removed. I will just point to the --help message.

tools/sched_ext/scx_layered/src/bpf/layered.h
htejun: So, that's what I had but it got really confusing too because `ors` is composed of `ands` instead of, well, `or`s. I'll see if there's some better naming convention.

tools/sched_ext/scx_layered/src/bpf/layered.bpf.c
htejun: It's a bit of chicken and egg because we can't really establish a stable state before actually running the logic and running the logic depends on bpf data structures being accessible. We can seed the cpumasks with something more reasonable. e.g. Just setting all bits is probably better.

htejun: Yeah, lemme just assert all bits and add a comment. This is just bootstrapping and stable state isn't established at all, so as long as the initial behavior isn't crazy, I'd rather not complicate code.

tools/sched_ext/scx_layered/src/bpf/layered.bpf.c
htejun: Hmm.. we still the cpumask refreshing. Maybe I can move that to PROG_RUN but I'm not sure that's supported in libbpf-rs. I'll switch it to timer.

Decave (approved):
LGTM, will let you merge once you've applied whatever changes you need.

htejun:
I kept the tick for cpumask updates instead of switching to timer because I didn't want to add an 1ms timer for this and inlining the update to hotpath would require more massaging to avoid concurrent updates.

The ands/ors naming, I just ran out of steam. I might take another look tomorrow but prepping the next iteration is likely higher priority, so maybe later. Let's merge it as-is and revisit these issues later.

PR #66, opened 2023-11-01
scx: Print scheduler state in panic message

Decave:
It would be useful to see what the sched_ext scheduler state is, and what scheduler is running, at panic time. Let's add a hook and print that information.

### DIFF
--- a/include/linux/sched/ext.h
+++ b/include/linux/sched/ext.h
@@ -705,10 +705,12 @@ struct sched_ext_entity {
 };
 
 void sched_ext_free(struct task_struct *p);
+void print_scx_info(const char *log_lvl, struct task_struct *p);
 
 #else	/* !CONFIG_SCHED_CLASS_EXT */
 
 static inline void sched_ext_free(struct task_struct *p) {}
+static inline void print_scx_info(const char *log_lvl, struct task_struct *p) {}
 
 #endif	/* CONFIG_SCHED_CLASS_EXT */
 #endif	/* _LINUX_SCHED_EXT_H */
--- a/kernel/sched/ext.c
+++ b/kernel/sched/ext.c
@@ -3688,6 +3688,68 @@ static void kick_cpus_irq_workfn(struct irq_work *irq_work)
 	cpumask_clear(this_rq->scx.cpus_to_wait);
 }
 
+/**
+ * print_scx_info - print out sched_ext scheduler state
+ * @log_lvl: the log level to use when printing
+ * @p: target task
+ *
+ * If @task is running on a sched_ext scheduler, print out the name of the
+ * sched_ext scheduler, and other various scheduler-related debugging
+ * information about the task.
+ *
+ * This function can be safely called on any task as long as the
+ * task_struct itself is accessible. While safe, this function isn't
+ * synchronized and may print out mixups or garbages of limited length.
+ */
+void print_scx_info(const char *log_lvl, struct task_struct *p)
+{
+	struct sched_class *class = NULL;
+	enum scx_ops_enable_state state = scx_ops_enable_state();
+	s64 delta = 0;
+	long ops_state = 0;
+	int task_cpu;
+	struct thread_info *thread_info;
+	const char *all = READ_ONCE(scx_switching_all) ? "+all" : "";
+
+	if (!scx_enabled() || state == SCX_OPS_DISABLED)
+		return;
+
+	/*
+	 * Carefully check if the task was running on sched_ext, and then
+	 * carefully copy the time it's been runnable, and its state.
+	 */
+	copy_from_kernel_nofault(&class, &p->sched_class, sizeof(class));
+	if (!class || class != &ext_sched_class) {
+		printk("%ssched_ext: %s (%s%s)", log_lvl, scx_ops.name,
+		       scx_ops_enable_state_str[state], all);
+		return;
+	}
+
+	copy_from_kernel_nofault(&thread_info, task_thread_info(p),
+				 sizeof(thread_info));
+	copy_from_kernel_nofault(&task_cpu, &thread_info->cpu,
+				 sizeof(task_cpu));
+	if (ops_cpu_valid(task_cpu)) {
+		struct rq *task_rq;
+		u64 rq_clock;
+		unsigned long runnable_at;
+
+		task_rq = cpu_rq(task_cpu);
+		copy_from_kernel_nofault(&rq_clock, &task_rq->clock,
+					 sizeof(rq_clock));
+		copy_from_kernel_nofault(&ops_state, &p->scx.ops_state.counter,
+					 sizeof(ops_state));
+		copy_from_kernel_nofault(&runnable_at, &p->scx.runnable_at,
+					 sizeof(runnable_at));
+		delta = rq_clock - runnable_at;
+	}
+
+	/* Print everything onto one line to conserve console spce. */
+	printk("%ssched_ext: %s (%s%s), task: runnable_at=%+lld state=%#lx",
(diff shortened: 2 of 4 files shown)


ByteLab-David:
@htejun updated this to only print when the scheduler is enabled, and allowed the behavior to be toggled with a default-off Kconfig option.

htejun:
This probably shouldn't be sth configurable as that can lead to a lot of confusion. Also, I'm not sure this is panic is the right hook up point. e.g. showing which scheduler is running can be very useful when hung task warning is triggered. So, it'd be better to be a part of basic cpu state dump which gets included whenever we printout backtrace.

htejun:
For an example, take a look at how print_worker_info() is used.

include/linux/sched/ext.h
```diff
@@ -705,10 +705,12 @@ struct sched_ext_entity {
 };
 
 void sched_ext_free(struct task_struct *p);
+void print_scx_info(const char *log_lvl, struct task_struct *p);
 
 #else	/* !CONFIG_SCHED_CLASS_EXT */
 
 static inline void sched_ext_free(struct task_struct *p) {}
+static inline void print_scx_info(const char *log_lvl, struct task_struct *p);
```
htejun: empty body missing?

kernel/sched/ext.c
```diff
@@ -3688,6 +3688,63 @@ static void kick_cpus_irq_workfn(struct irq_work *irq_work)
...
+ * print_scx_info - print out sched_ext scheduler state
+ * @log_lvl: the log level to use when printing
+ * @p: target task
+ *
+ * If @task is running on a sched_ext scheduler, print out the name of the
+ * sched_ext scheduler, and other various scheduler-related debugging
+ * information about the task.
+ *
+ * This function can be safely called on any task as long as the
+ * task_struct itself is accessible.  While safe, this function isn't
```
htejun: Single space instead of double for consistency?

kernel/sched/ext.c
```diff
@@ -3688,6 +3688,63 @@ static void kick_cpus_irq_workfn(struct irq_work *irq_work)
...
+	if (!scx_enabled() || state == SCX_OPS_DISABLED)
+		return;
+
+	/*
+	 * Carefully check if the task was running on sched_ext, and then
+	 * carefully copy the time it's been runnable, and its state.
+	 */
+	copy_from_kernel_nofault(&class, &p->sched_class, sizeof(class));
+	task_scx_enabled = class && class == &ext_sched_class;
+	if (task_scx_enabled) {
```
htejun: This really should be gated on "is scx enabled on the system at all?" rather than "is this task on scx?"

kernel/sched/ext.c
```diff
@@ -3688,6 +3688,63 @@ static void kick_cpus_irq_workfn(struct irq_work *irq_work)
...
+		copy_from_kernel_nofault(&runnable_at, &p->scx.runnable_at,
+					 sizeof(runnable_at));
+		copy_from_kernel_nofault(&thread_info, task_thread_info(p),
+					 sizeof(thread_info));
+		copy_from_kernel_nofault(&task_cpu, &thread_info->cpu,
+					 sizeof(task_cpu));
+		task_rq = cpu_rq(task_cpu);
```
htejun, line 3737: Prolly wanna test task_cpu is actually valid before cpu_rq().

kernel/sched/ext.c
```diff
@@ -3688,6 +3688,63 @@ static void kick_cpus_irq_workfn(struct irq_work *irq_work)
...
+		copy_from_kernel_nofault(&rq_clock, &task_rq->clock, sizeof(rq_clock));
+		copy_from_kernel_nofault(&ops_state, &p->scx.ops_state.counter,
+					 sizeof(ops_state));
+		if (rq_clock >= runnable_at)
+			delta = rq_clock - runnable_at;
```
htejun: Negative delta is great too.

kernel/sched/ext.c
```diff
@@ -3688,6 +3688,63 @@ static void kick_cpus_irq_workfn(struct irq_work *irq_work)
...
+	}
+	/* Print everything onto one line to conserve console spce. */
+	printk("%ssched_ext: ops: %s, state: %s, switching all: %d, "
+	       "task: scx enabled: %d, runnable time: %llu, ops_state: %lx" ,
```
htejun: How about sth like the following?
`sched_ext: $NAME ($STATE[+all])` if task is not on scx.
`sched_ext: $NAME ($STATE[+all]), task: runnable_at=[+-]DELTA state=$OPS_STATE` if on scx

htejun (changes requested):

kernel/sched/ext.c
ByteLab-David: I think the `if (!scx_enabled()` above should address that?

kernel/sched/ext.c
ByteLab-David: Not sure if `ops_cpu_valid()` is kosher to use here? Seems like exactly what we want though

include/linux/sched/ext.h
ByteLab-David: This made me realize !EXT builds are broken in general. I'll send a separate PR to fix

kernel/sched/ext.c
ByteLab-David: Not 100% sure if this is what you wanted. Feel free to clarify

kernel/sched/ext.c
```diff
@@ -3688,6 +3688,68 @@ static void kick_cpus_irq_workfn(struct irq_work *irq_work)
...
+{
+	struct sched_class *class = NULL;
+	unsigned long runnable_at = 0;
+	u64 rq_clock = 0;
+	s64 delta = 0;
+	long ops_state = 0;
+	int task_cpu = 0;
+	struct thread_info *thread_info = NULL;
+	struct rq *task_rq = NULL;
+	enum scx_ops_enable_state state = scx_ops_enable_state();
```
htejun: Let's move decls into the blocks that they're used.

kernel/sched/ext.c
```diff
@@ -3688,6 +3688,68 @@ static void kick_cpus_irq_workfn(struct irq_work *irq_work)
...
+			copy_from_kernel_nofault(&ops_state,
+						 &p->scx.ops_state.counter,
+						 sizeof(ops_state));
+			delta = rq_clock - runnable_at;
+		}
+
+		/* Print everything onto one line to conserve console spce. */
+		printk("%ssched_ext: %s (%s[%d]), task: runnable_since=%lld state=%lx",
+		       log_lvl, scx_ops.name, scx_ops_enable_state_str[state],
+		       READ_ONCE(scx_switching_all), delta, ops_state);
```
htejun: For all, I meant `"...(%s%s)...", all ? "+all" : ""`. Also, we prolly want `%+lld` for runnable timestamp to signify that it's a delta value and `0x%lx` for state.

kernel/sched/ext.c
```diff
@@ -3688,6 +3688,68 @@ static void kick_cpus_irq_workfn(struct irq_work *irq_work)
...
+	} else {
+		printk("%ssched_ext: %s (%s[%d])", log_lvl, scx_ops.name,
+		       scx_ops_enable_state_str[state],
+		       READ_ONCE(scx_switching_all));
+	}
+	pr_cont("\n");
```
htejun: Let's avoid pr_cont. When multiple cpus are printing, the msg fragments get all mixed up.

htejun (changes requested):

htejun (approved):

PR #70, opened 2023-11-04
scx_rusty: Usage ravg for dom and task loads

htejun:
One weakness of rusty is that its CPU usage can become substantial on systems with large number of threads as it had to iterate the task_ctxs for all tasks on every LB period to calculate task loads and per-domain sums. This pull request switches scx_rusty to use ravg instead.

- Per-domain load sums and per-task duty cycles are tracked using ravg. Task load is then calculated by multiplying its weight to the duty cycle. This is because duty cycle will later be useful for solving infeasible weights.
- Userspace monitors per-domain load every period but loads task loads only when necessary. Each domain tracks the latest 1k active tasks and only the active tasks which belong to the pushing domains are loaded. In most cases, this should lead to negligible CPU usage even on fairly active systems.


tools/sched_ext/ravg_impl.bpf.h
```diff
@@ -0,0 +1,359 @@
+/* to be included in the main bpf.c file */
+#include "ravg.bpf.h"
+
+#define RAVG_FN_ATTRS		inline __attribute__((unused, always_inline))
+//#define RAVG_FN_ATTRS		__attribute__((unused))
```
ByteLab-David: extra comment

Decave (approved):
Looks good after skimming. Feel free to land after fixing that nit about the comment, but it might be a good idea to also get a look over from Dan.

tools/sched_ext/ravg_impl.bpf.h
```diff
@@ -0,0 +1,359 @@
...
+
+static RAVG_FN_ATTRS u64 ravg_decay(u64 v, u32 shift)
+{
+	if (shift >= 64)
+		return 0;
+	else
+		return v >> shift;
+}
+
+static RAVG_FN_ATTRS u32 ravg_normalize_dur(u32 dur, u32 half_life)
```
dschatzberg, line 24: It's not really clear to me what this function is doing. Can you elaborate a bit?

tools/sched_ext/ravg_impl.bpf.h
```diff
@@ -0,0 +1,359 @@
...
+ * subtracted if @is_fer_in is %false and added if %true.
+ */
+static RAVG_FN_ATTRS void ravg_transfer(struct ravg_data *base, u64 base_new_val,
+					struct ravg_data *xfer, u64 xfer_new_val,
+					u32 half_life, bool is_xfer_in)
+{
+	/* synchronize @base and @xfer */
+	if ((s64)(base->val_at - xfer->val_at) < 0)
+		ravg_accumulate(base, base_new_val, xfer->val_at, half_life);
+	else if ((s64)(base->val_at - xfer->val_at) > 0)
```
dschatzberg, line 167: These comparisons to 0 read a little oddly to me - isn't it the same to just compare `base->val_at < xfer->val_at`?

tools/sched_ext/scx_rusty/src/bpf/rusty.bpf.c
```diff
@@ -95,7 +98,133 @@ struct {
...
+	lockw = bpf_map_lookup_elem(&dom_load_locks, &dom_id);
+
+	if (!domc || !lockw) {
+		scx_bpf_error("dom_ctx / lock lookup failed");
+		return;
+	}
+
+	bpf_spin_lock(&lockw->lock);
+	domc->load += adj;
+	ravg_accumulate(&domc->load_rd, domc->load, now, load_half_life);
```
dschatzberg, line 141: I'm confused by `domc->load += adj` and it makes me realize I didn't understand this ravg_accumulate interface - why aren't you just accumulating adj directly?

tools/sched_ext/scx_rusty/src/bpf/rusty.bpf.c
```diff
@@ -95,7 +98,133 @@ struct {
...
+	task_load_rd = taskc->dcyc_rd;
+	ravg_scale(&task_load_rd, p->scx.weight, 0);
+
+	if (debug >= 2)
+		task_load = ravg_read(&task_load_rd, now, load_half_life);
+
+	/* transfer out of @from_dom_id */
+	bpf_spin_lock(&from_lockw->lock);
+	if (taskc->runnable)
+		from_domc->load -= p->scx.weight;
```
dschatzberg, line 190: Why are you subtracting weight here? I guess the assumption is if its runnable then it has full duty cycle?

dschatzberg (commented):
This looks good - just a few questions inline where I wasn't clear on the behavior

tools/sched_ext/ravg_impl.bpf.h
htejun: It's just converting @dur into [0, 1] fraction of @half_life. So, if half_life is 2s and dur is 0.4s, the returned value is 0.2. This can be used when e.g. blending rd->old and ->cur. If we're currently at 20% through the current half life period, the returned result is the blend of 80% old and 20% cur.

tools/sched_ext/ravg_impl.bpf.h
htejun: So, imagine the whole value range mapped to a circle where it winds all the way to all 1 bits and then to zero and you want to decide whether one value is before or after the other. The absoulte value doesn't matter but just the relative positions. Let's say A and B. A is at some value and the value circle can be divided into two halves - one closer to A's forward direction, the other closer to A's backward direction. If B is in A's forward direction half, it's ahead. If B is in A's backward direction half, it's behind. ie. it's a way to establish ordering regardless of the actual value overflowing as long as the distance between A and B is less than half of the full value range.

This is really important when using 32bits to track high granularity time stamps but much less so with 64bit. With nsec ktime, this likely isn't a concern but w/ 64bit jiffies, kernel intentionally makes this a concern by setting epoch to a very high value which overflows soon after booting.

So, technically, we'll be okay with simple comparisons, but idk it's a habit I've developed over time and probably not a bad one to have when dealing with timestampy values.

tools/sched_ext/scx_rusty/src/bpf/rusty.bpf.c
htejun: At a given point in time, there are some number of tasks bound to a domain. When such a task becomes runnable, it contributes to the load - the active weight sum - of the domain. e.g. if there are three runnable weight 100 tasks, the current load of the domain is 300. The load avg of the domain is determined as the running avg of this load number.

So, domc->load is tracking the sum of currently runnable weights in the domain and domc->load_rd is tracking the running average of that value. So, when a task becomes runnable, we add its weight to domc->load and tell ravg to 1. integrate the previous value between the previous accumulation call and now and 2. set the current value to the newly updated domc->load so that it can be integrated to the period between now and the next call.

tools/sched_ext/scx_rusty/src/bpf/rusty.bpf.c
htejun: domc->load doesn't have much to do with duty cycle. It's just the current intantenous sum of runnable weights. When ravg takes runnable average of domc->load, that ends up factoring in the duty cycles of the different weights because that's captured in the changes of intanteneous runnable weight sum over time.

When a task migrates from one domain to another, we need to update both 1. the instantenous sums of the two domains if the task is runnable 2. the task's contribution to the running avgs of the instantenous sums. So, the above part is doing 1. If the task is currently runnable, it was contributing to the current instantenous sum of the from domain, so we subtract that and then move it to the to_domc.

One can argue whether 2 is necessary or not - e.g. let's say from_dom has load_avg of 100, to_dom 50 and the migrating task 50. After the task moves, if we move the running avg contribution together, from_dom's load_avg will become 50 and to_dom 100 right away. If we don't, they will converge to those values over time. While the latter behavior can be argued to be valid in some sense, it ends up interfereing with subsequent load balance rounds because the load which has already been moved would still be contributing to the load avgs. So, we should move the contributions together.

PR #92, opened 2023-12-04
Change semantics of FIFO/PRIQ dispatching

Decave:


htejun (approved):
Do you mind updating the docs too?

Decave:
Following our offline convo I realized that the docs already say this:

>When a CPU is looking for the next task to run, if the local DSQ is not empty, the first task is picked. Otherwise, the CPU tries to consume the global DSQ. If that doesn't yield a runnable task either, ops.dispatch() is invoked.

Let me know if that's OK or if I should update further.

ByteLab-David:
Done, I realized that the docs already had a description of consume order (https://github.com/sched-ext/sched_ext/blob/nest_scheduler/Documentation/scheduler/sched-ext.rst?plain=1#L122-L125), but I added some documentation mentioning that you can't do vtime dispatching to internal DSQs.

PR #100, opened 2023-12-14
Fix and update semantics for ops.enable() and ops.disable()

Decave:
- A task's scx.weight is properly set before ops.enable() on the fork path, but not on the BPF prog load path. Set it properly for both.
- ops.enable() (and a corresponding ops.disable()) is called once for every task on the system on the BPF prog load path, or the fork path. It would be useful to instead call ops.enable() whenever a task is about to run on scx, and ops.disable() whenever a task is no longer running on scx.

### DIFF
--- a/include/linux/sched/ext.h
+++ b/include/linux/sched/ext.h
@@ -693,6 +693,7 @@ struct sched_ext_entity {
 #ifdef CONFIG_SCHED_CORE
 	u64			core_sched_at;	/* see scx_prio_less() */
 #endif
+	u64			ddsq_id;
 
 	/* BPF scheduler modifiable fields */
 
@@ -716,12 +717,6 @@ struct sched_ext_entity {
 	 * recommended.
 	 */
 	u64			dsq_vtime;
-
-	/*
-	 * Used to track when a task has requested a direct dispatch from the
-	 * ops.select_cpu() path.
-	 */
-	u64			ddsq_id;
 
 	/*
 	 * If set, reject future sched_setscheduler(2) calls updating the policy
--- a/include/linux/sched/ext.h
+++ b/include/linux/sched/ext.h
@@ -431,11 +431,11 @@ struct sched_ext_ops {
 	 *
 	 * Either we're loading a BPF scheduler or a new task is being forked.
 	 * Prepare BPF scheduling for @p. This operation may block and can be
-	 * used for allocations.
+	 * used for allocations, and is called exactly once for a task.
 	 *
 	 * Return 0 for success, -errno for failure. An error return while
-	 * loading will abort loading of the BPF scheduler. During a fork, will
-	 * abort the specific fork.
+	 * loading will abort loading of the BPF scheduler. During a fork, it
+	 * will abort that specific fork.
 	 */
 	s32 (*prep_enable)(struct task_struct *p, struct scx_enable_args *args);
 
@@ -444,8 +444,9 @@ struct sched_ext_ops {
 	 * @p: task to enable BPF scheduling for
 	 * @args: enable arguments, see the struct definition
 	 *
-	 * Enable @p for BPF scheduling. @p is now in the cgroup specified for
-	 * the preceding prep_enable() and will start running soon.
+	 * Enable @p for BPF scheduling. @p is now in the cgroup specified in
+	 * @args. enable() is called on @p any time it enters SCX, and is
+	 * always paired with a matching disable().
 	 */
 	void (*enable)(struct task_struct *p, struct scx_enable_args *args);
 
@@ -465,7 +466,8 @@ struct sched_ext_ops {
 	 * @p: task to disable BPF scheduling for
 	 *
 	 * @p is exiting, leaving SCX or the BPF scheduler is being unloaded.
-	 * Disable BPF scheduling for @p.
+	 * Disable BPF scheduling for @p. A disable() call is always matched
+	 * with a prior enable() call.
 	 */
 	void (*disable)(struct task_struct *p);
 
@@ -606,14 +608,15 @@ enum scx_ent_flags {
 	SCX_TASK_QUEUED		= 1 << 0, /* on ext runqueue */
 	SCX_TASK_BAL_KEEP	= 1 << 1, /* balance decided to keep current */
 	SCX_TASK_DDSP_PRIQ	= 1 << 2, /* task should be enqueued on priq when directly dispatched */
-
-	SCX_TASK_OPS_PREPPED	= 1 << 8, /* prepared for BPF scheduler enable */
-	SCX_TASK_OPS_ENABLED	= 1 << 9, /* task has BPF scheduler enabled */
+	SCX_TASK_STATE_0	= 1 << 3, /* first bit encoding the task's current state */
+	SCX_TASK_STATE_1	= 1 << 4, /* second bit encoding the task's current state */
 
 	SCX_TASK_WATCHDOG_RESET = 1 << 16, /* task watchdog counter should be reset */
 	SCX_TASK_DEQD_FOR_SLEEP	= 1 << 17, /* last dequeue was for SLEEP */
 
 	SCX_TASK_CURSOR		= 1 << 31, /* iteration cursor, not a task */
+
+	SCX_TASK_STATE_MASK	= SCX_TASK_STATE_0 | SCX_TASK_STATE_1,
 };
 
 /* scx_entity.dsq_flags */
@@ -645,6 +648,21 @@ enum scx_kf_mask {
 	__SCX_KF_RQ_LOCKED	= SCX_KF_CPU_RELEASE | SCX_KF_DISPATCH |
 				  SCX_KF_ENQUEUE | SCX_KF_SELECT_CPU | SCX_KF_REST,
 	__SCX_KF_TERMINAL	= SCX_KF_ENQUEUE | SCX_KF_SELECT_CPU | SCX_KF_REST,
+};
+
+/* scx_entity.task_state */
+enum scx_task_state {
+	/* ops.prep_enable() has not yet been called on task */
+	SCX_TASK_NONE,
+
+	/* ops.prep_enable() succeeded on task, but it still be cancelled */
+	SCX_TASK_INIT,
+
+	/* Task is fully initialized, but not being scheduled in sched_ext */
+	SCX_TASK_READY,
+
+	/* Task is fully initialized and is being scheduled in sched_ext */
+	SCX_TASK_ENABLED,
 };
 
 /*
--- a/include/linux/sched/ext.h
+++ b/include/linux/sched/ext.h
@@ -123,12 +123,18 @@ enum scx_ops_flags {
 				  SCX_OPS_CGROUP_KNOB_WEIGHT,
 };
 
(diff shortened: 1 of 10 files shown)


include/linux/sched/ext.h
```diff
@@ -417,47 +417,59 @@ struct sched_ext_ops {
...
+	 * @args: init arguments, see the struct definition
 	 *
-	 * Enable @p for BPF scheduling. @p is now in the cgroup specified for
-	 * the preceding prep_enable() and will start running soon.
+	 * @p was init_task()'d but failed before reaching enable(). Undo the
+	 * preparation.
 	 */
-	void (*enable)(struct task_struct *p, struct scx_enable_args *args);
+	void (*cancel_init)(struct task_struct *p,
+			    struct scx_enable_args *args);
```
htejun: As discussed in the chat, it'd be great if we can just do four methods - init/exit_task() and enable/disable(). We'll probably want to add an argument to exit_task() to indicate whether it's being called to cancel init or it's for a task exit.

kernel/sched/ext.c
```diff
@@ -2274,69 +2274,74 @@ static int scx_ops_prepare_task(struct task_struct *p, struct task_group *tg)
...
-	p->scx.flags &= ~SCX_TASK_OPS_PREPPED;
-	p->scx.flags |= SCX_TASK_OPS_ENABLED;
+	p->scx.flags |= SCX_TASK_OPS_LOADED;
+
+	if (SCX_HAS_OP(set_weight))
+		SCX_CALL_OP_TASK(SCX_KF_REST, set_weight, p, p->scx.weight);
 }
 
-static void scx_ops_disable_task(struct task_struct *p)
+static void scx_ops_disable_task(struct task_struct *p, bool transitioning)
```
htejun: I think it'd be cleaner if we separate out disabling and exiting into separate functions. If you look at the body of the function, they're pretty separate paths anyway.

kernel/sched/ext.c
```diff
@@ -2237,18 +2237,18 @@ static int scx_ops_prepare_task(struct task_struct *p, struct task_group *tg)
 {
```
htejun, line 2330: Maybe match the function name so that it's `scx_ops_init_task()`?

kernel/sched/ext.c
```diff
@@ -2361,21 +2366,20 @@ int scx_fork(struct task_struct *p)
 void scx_post_fork(struct task_struct *p)
 {
 	if (scx_enabled()) {
-		struct rq_flags rf;
-		struct rq *rq;
-
-		rq = task_rq_lock(p, &rf);
+		p->scx.flags |= SCX_TASK_OPS_LOADABLE;
```
htejun: I don't understand what `LOADABLE` is doing. What's its role?

kernel/sched/ext.c
```diff
@@ -2404,19 +2410,24 @@ void sched_ext_free(struct task_struct *p)
 	 * @p is off scx_tasks and wholly ours. scx_ops_enable()'s PREPPED ->
 	 * ENABLED transitions can't race us. Disable ops for @p.
 	 */
-	if (p->scx.flags & (SCX_TASK_OPS_PREPPED | SCX_TASK_OPS_ENABLED)) {
+	if (p->scx.flags &
+	    (SCX_TASK_OPS_INITIALIZED | SCX_TASK_OPS_LOADABLE)) {
```
htejun: Can we do `INITED` and `ENABLED`?

htejun:
So, one thing which is weird is `scx_enable_args.cgroup`. This is provided because the associated cgroup is not available at the time of `.prep_enable()` and it just carries over to `.enable()` because they're considered two continued parts of the same operation. After this change, `init_task` and `enable` are distinct steps and while `init_task` would need the explicit cgroup argument, `enable` won't. So, I guess what needs to happen is introducing `scx_init_task_args` and remove cgroup from` scx_enable_args`?

kernel/sched/ext.c
ByteLab-David: `LOADABLE` reflects a task with the following properties:
1. It has been initialized with `ops.init_task()`
2. It can no longer be cancelled with `ops.cancel_init()`
3. It is _not_ actively enabled.

In other words, it's there to tell us when to issue `ops.cancel_init()` rather than `ops.exit_task()`. `ops.cancel_init()` is invoked when we want to notify the scheduler that a task that was previously initialized with `ops.init_task()` was never actually fully initialized. In this case we would have `(p->scx.flags & SCX_TASK_OPS_INITIALIZED) != 0`, but `(p->scx.flags & SCX_TASK_OPS_LOADED) == 0`. We need this flag in addition to `SCX_TASK_OPS_LOADED` so we know whether or not the task has been enabled, and thus whether we should call `ops.disable()` when the task is exiting.

Edit: Note that getting rid of `ops.cancel_init()` doesn't really fix the problem because we still have to know how to invoke `ops.exit_task()`.

kernel/sched/ext.c
ByteLab-David: See the comment above r.e. LOADABLE -- the problem is that we need that flag to know whether a task was ever _fully_ initialized. If we used `ENABLED`, I think we'd need something like `ENABLABLE` for that flag if we wanted to be consistent.

kernel/sched/ext.c
ByteLab-David: Hmm, but we have to invoke `ops.disable()` on the exit path as well, no?

htejun: Ah, right, the branch is between `.disable` and `.cancel_init` which should go away if we get rid of `.cancel_init`.

ByteLab-David: Not sure I'm 100% following. I think we'd still have that branch even if we got rid of `.cancel_init()` because we would have to know which arg we should pass to `.exit_task()`, right?

ByteLab-David:
> So, one thing which is weird is `scx_enable_args.cgroup`. This is provided because the associated cgroup is not available at the time of `.prep_enable()` and it just carries over to `.enable()` because they're considered two continued parts of the same operation. After this change, `init_task` and `enable` are distinct steps and while `init_task` would need the explicit cgroup argument, `enable` won't. So, I guess what needs to happen is introducing `scx_init_task_args` and remove cgroup from` scx_enable_args`?

@htejun so I applied most of your suggestions and pushed. The ones I haven't applied are:
- Splitting disable and exit into separate paths, as we can't get rid of that branch for the `ops.cancel_init()` path (even though it's now `ops.exit_task()` as we have to know when a task failed to be fully initialized.
- Changing `SCX_TASK_OPS_LOADED` to `SCX_TASK_OPS_ENABLED`, as I think we'd also have to `SCX_OPS_TASK_LOADABLE` to `SCX_TASK_OPS_ENABLABLE` to be consistent. I'm OK with applying that if you want but it's obviously a pretty weird word so IMO it doesn't do the readability any favors.
- I updated `ops.enable()` to not take an arg, as now it doesn't actually receive anything. I'm happy to make empty struct args to `ops.enable()` and `ops.disable()` for future proofing, though we should probably be consistent and apply that to other callbacks too like `ops.init()` if we decide to go that route. IMO it would be better to keep the args minimal until we know we need them, assuming we can rely on some libbpf magic to help us with resolving the correct struct_ops map at load time.

I'm happy to revisit any of the above, just let me know.

htejun (changes requested):
I'm afraid I don't like the code. It's verbose and confusing at the same time, which isn't a good combination and things like `if (loadable || loaded)` are actively confusing - ie. does that mean something loaded may not be loadable? Why is it loaded and loadable when none of the verbs are load? Why do we still need `.cancel_init`? Can we just not distinguish `loadable`? Or at least make it a flag which indicates `has ever been enabled` which doesn't actively participate in state transitions and just used in a specific spot to tell `.exit_task` what's going on?

I like the eventual behavior this PR is targeting but there gotta be a more intuitive way to implement this.

ByteLab-David:
No problem, let me take a stab at cleaning this up a bit

ByteLab-David:
@htejun this should be ready for another review. Thanks for pushing back on the last iteration -- I think this one is way cleaner and more readable.

include/linux/sched/ext.h
```diff
@@ -647,6 +646,21 @@ enum scx_kf_mask {
...
+
+	/* ops.prep_enable() succeeded on task, but it still be cancelled */
+	SCX_TASK_PRE_INITIALIZED,
+
+	/* Task is fully initialized, but not being scheduled in sched_ext */
+	SCX_TASK_INITIALIZED,
+
+	/* Task is fully initialized and is being scheduled in sched_ext */
+	SCX_TASK_ENABLED,
+};
```
htejun, line 671: - Can we shorter names? e.g. NONE, INITING, INITED, ENABLED. If that's too ugly, how about NONE, INIT, READY, ENABLED?
- Can we put this into two bits in the flags field and have get/set_task_state helpers? The sanity checks on allowed state transitions can be moved into the set helper.

kernel/sched/ext.c
```diff
@@ -3445,15 +3462,15 @@ static int scx_ops_enable(struct sched_ext_ops *ops)
 			sched_deq_and_put_task(p, DEQUEUE_SAVE | DEQUEUE_MOVE,
 					       &ctx);
 
-			scx_ops_enable_task(p);
+			p->scx.task_state = SCX_TASK_INITIALIZED;
```
htejun: The comment block above says PREPPED -> ENABLED. Maybe update together?

include/linux/sched/ext.h
```diff
@@ -123,14 +123,28 @@ enum scx_ops_flags {
...
+	SCX_TASK_EXIT_NORMAL,
+
+	/* Task exited after being initialized, but before being enabled. */
+	SCX_TASK_EXIT_CANCELLED,
+};
+
+/* argument container for ops.exit_task() */
+struct scx_exit_task_args {
+	/* %SCX_TASK_EXIT_* - broad category of the task exit reason */
+	enum scx_exit_task_kind	kind;
```
htejun: Can this just be a bool? I have a hard time imagining what other exit kinds there can be.

htejun (approved):
I left a few comments but generally looks good to me. Thanks a lot for reworking it.

include/linux/sched/ext.h
ByteLab-David: So one example I was considering is if the task is cancelled because it's TASK_DEAD (see the `if (READ_ONCE(p->__state) != TASK_DEAD)` check in `scx_ops_enable()`) rather than because another task failed to be initialized in `ops.init_task()`. In that case the kind would maybe be something like `SCX_TASK_EXIT_DYING`. I can't think of any others, so if we definitely don't want to distinguish there we can make it a `bool`.

include/linux/sched/ext.h
ByteLab-David: >how about NONE, INIT, READY, ENABLED?

will do

>Can we put this into two bits in the flags field and have get/set_task_state helpers? The sanity checks on allowed state transitions can be moved into the set helper.

hah, cheeky. sure, will do.

include/linux/sched/ext.h
ByteLab-David: I'll just make it a bool. We can always change it to an enum if we need later

PR #103, opened 2024-01-02
Implement generic bypass mode and use it while PM operations are in progress

htejun:
#102 reports that `scx_userland` and `scx_rustland` trigger stall failure on system suspend attempts. The problem is that these schedulers need userspace running for scheduling to work. Because during suspend and other PM operations, all user and some kernel threads are frozen during preparation, scheduling stops working triggering either freezer timeout or scx watchdog.

In #102, @arighi worked around by preemptively ejecting the sched_ext scheduler. While it's possible to implement userspace hooks to reload the scheduler after PM operations are finished, it's rather inconvenient and unnecessary. In terms of scheduling, there isn't anything interesting happening while the PM operations are in progress, so all we need to do is not stalling the system.

In the ops disabling path, we're already doing something similar. Because we can't trust the scheduler, we override SCX into global FIFO mode before taking any actions which may require working scheduling. This PR

- Factors out the forced FIFO behavior in the disabling path and generalizes it into bypass mode.
- Optimize bypass mode so that it doesn't involve iterating all tasks in the system.
- Enables bypass mode while PM operations are in progress so that the system guarantees forward progress regardless of whether the SCX scheduler depends on userspace or not.

### DIFF
--- a/kernel/sched/ext.c
+++ b/kernel/sched/ext.c
@@ -946,25 +946,25 @@ global:
 	dispatch_enqueue(&scx_dsq_global, p, enq_flags);
 }
 
-static bool watchdog_task_watched(const struct task_struct *p)
+static bool task_runnable(const struct task_struct *p)
 {
-	return !list_empty(&p->scx.watchdog_node);
+	return !list_empty(&p->scx.runnable_node);
 }
 
-static void watchdog_watch_task(struct rq *rq, struct task_struct *p)
+static void set_task_runnable(struct rq *rq, struct task_struct *p)
 {
 	lockdep_assert_rq_held(rq);
-	if (p->scx.flags & SCX_TASK_WATCHDOG_RESET)
+	if (p->scx.flags & SCX_TASK_RESET_RUNNABLE_AT)
 		p->scx.runnable_at = jiffies;
-	p->scx.flags &= ~SCX_TASK_WATCHDOG_RESET;
-	list_add_tail(&p->scx.watchdog_node, &rq->scx.watchdog_list);
+	p->scx.flags &= ~SCX_TASK_RESET_RUNNABLE_AT;
+	list_add_tail(&p->scx.runnable_node, &rq->scx.runnable_list);
 }
 
-static void watchdog_unwatch_task(struct task_struct *p, bool reset_timeout)
+static void clr_task_runnable(struct task_struct *p, bool reset_runnable_at)
 {
-	list_del_init(&p->scx.watchdog_node);
-	if (reset_timeout)
-		p->scx.flags |= SCX_TASK_WATCHDOG_RESET;
+	list_del_init(&p->scx.runnable_node);
+	if (reset_runnable_at)
+		p->scx.flags |= SCX_TASK_RESET_RUNNABLE_AT;
 }
 
 static void enqueue_task_scx(struct rq *rq, struct task_struct *p, int enq_flags)
@@ -986,11 +986,11 @@ static void enqueue_task_scx(struct rq *rq, struct task_struct *p, int enq_flags
 		sticky_cpu = cpu_of(rq);
 
 	if (p->scx.flags & SCX_TASK_QUEUED) {
-		WARN_ON_ONCE(!watchdog_task_watched(p));
+		WARN_ON_ONCE(!task_runnable(p));
 		return;
 	}
 
-	watchdog_watch_task(rq, p);
+	set_task_runnable(rq, p);
 	p->scx.flags |= SCX_TASK_QUEUED;
 	rq->scx.nr_running++;
 	add_nr_running(rq, 1);
@@ -1008,7 +1008,8 @@ static void ops_dequeue(struct task_struct *p, u64 deq_flags)
 {
 	unsigned long opss;
 
-	watchdog_unwatch_task(p, false);
+	/* dequeue is always temporary, don't reset runnable_at */
+	clr_task_runnable(p, false);
 
 	/* acquire ensures that we see the preceding updates on QUEUED */
 	opss = atomic_long_read_acquire(&p->scx.ops_state);
@@ -1055,7 +1056,7 @@ static void dequeue_task_scx(struct rq *rq, struct task_struct *p, int deq_flags
 	struct scx_rq *scx_rq = &rq->scx;
 
 	if (!(p->scx.flags & SCX_TASK_QUEUED)) {
-		WARN_ON_ONCE(watchdog_task_watched(p));
+		WARN_ON_ONCE(task_runnable(p));
 		return;
 	}
 
@@ -1710,7 +1711,7 @@ static void set_next_task_scx(struct rq *rq, struct task_struct *p, bool first)
 	if (SCX_HAS_OP(running) && (p->scx.flags & SCX_TASK_QUEUED))
 		SCX_CALL_OP_TASK(SCX_KF_REST, running, p);
 
-	watchdog_unwatch_task(p, true);
+	clr_task_runnable(p, true);
 
 	/*
 	 * @p is getting newly scheduled or got kicked after someone updated its
@@ -1772,13 +1773,13 @@ static void put_prev_task_scx(struct rq *rq, struct task_struct *p)
 	 */
 	if (p->scx.flags & SCX_TASK_BAL_KEEP) {
 		p->scx.flags &= ~SCX_TASK_BAL_KEEP;
-		watchdog_watch_task(rq, p);
+		set_task_runnable(rq, p);
 		dispatch_enqueue(&rq->scx.local_dsq, p, SCX_ENQ_HEAD);
 		return;
 	}
 
 	if (p->scx.flags & SCX_TASK_QUEUED) {
-		watchdog_watch_task(rq, p);
+		set_task_runnable(rq, p);
 
 		/*
 		 * If @p has slice left and balance_scx() didn't tag it for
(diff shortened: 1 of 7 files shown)


arighi:
I did an initial test in a VM and scx_rustland seems solid now, it even survived a busy loop of multiple suspend/resume. But for some reasons (that I still don't understand) I can still trigger the watchdog failure with scx_userland sometimes (see the traces below).

I'll run some tests also on my laptop later today.

```
[   45.220420] sched_ext: BPF scheduler "userland" errored, disabling
[   45.220432] sched_ext: runnable task stall (watchdog failed to check in for 5.001s)
[   45.220434]    scheduler_tick+0x1e1/0x3d0
[   45.220441]    update_process_times+0x9c/0xb0
[   45.220444]    tick_sched_handle+0x33/0x50
[   45.220447]    tick_nohz_highres_handler+0x6f/0x90
[   45.220449]    __hrtimer_run_queues+0x1af/0x390
[   45.220451]    hrtimer_interrupt+0xf8/0x230
[   45.220453]    __sysvec_apic_timer_interrupt+0x64/0x1a0
[   45.220457]    sysvec_apic_timer_interrupt+0x3a/0x80
[   45.220462]    asm_sysvec_apic_timer_interrupt+0x1a/0x20

[   32.178332] sched_ext: BPF scheduler "userland" errored, disabling
[   32.178338] sched_ext: runnable task stall (watchdog failed to check in for 5.001s)
[   32.178339]    scheduler_tick+0x1e1/0x3d0
[   32.178348]    update_process_times+0x9c/0xb0
[   32.178352]    tick_sched_handle+0x33/0x50
[   32.178356]    tick_nohz_highres_handler+0x6f/0x90
[   32.178358]    __hrtimer_run_queues+0x1af/0x390
[   32.178360]    hrtimer_interrupt+0xf8/0x230
[   32.178362]    __sysvec_apic_timer_interrupt+0x64/0x1a0
[   32.178367]    sysvec_apic_timer_interrupt+0x6f/0x80
[   32.178373]    asm_sysvec_apic_timer_interrupt+0x1a/0x20
[   32.178377]    __schedule+0x413/0xea0
[   32.178380]    schedule_idle+0x2a/0x40
[   32.178381]    cpu_startup_entry+0x2a/0x30
[   32.178384]    start_secondary+0xfc/0x100
[   32.178386]    secondary_startup_64_no_verify+0x178/0x17b
```

arighi:
@htejun the issue with scx_userland might be related to userland itself, nothing to do with your patch I think. I was able to reproduce the stall in virtme-ng with userland even without suspending/hibernating at all (and also in a kernel without this patch applied).

I think scx_userland needs a better logic to synchronize the BPF and user-space part, something similar to what we're doing in scx_rustland (with the nr_queued / nr_scheduled counters) and the `.update_idle` callback.

ByteLab-David:
> @htejun the issue with scx_userland might be related to userland itself, nothing to do with your patch I think. I was able to reproduce the stall in virtme-ng with userland even without suspending/hibernating at all (and also in a kernel without this patch applied).
> 
> I think scx_userland needs a better logic to synchronize the BPF and user-space part, something similar to what we're doing in scx_rustland (with the nr_queued / nr_scheduled counters) and the `.update_idle` callback.

Yep, this makes sense. Let me send out a patch that adds this tracking. Did you do anything in particular to trigger the hang?

arighi:
> Yep, this makes sense. Let me send out a patch that adds this tracking. Did you do anything in particular to trigger the hang?

Nothing special, I just started the kernel inside virtme-ng, opened tmux with a `dmesg -w` session, a session with `scx_userland` and a bash session where I'm running commands (that I was supposed to use to `echo disk > /sys/power/state`, but I was able to trigger a stall just by running `ls`).

arighi:
Just tested on real hardware as well, `scx_rustland` survived suspend/resume cycle correctly, but after stopping  it and trying to load `scx_userland` I got the following error:
```
[  153.074727] sched_ext: ops.prep_enable() failed (-12) for kworker/R-rcu_p[5] while loading
[  153.082693] sched_ext: BPF scheduler "userland" errored, disabling
```

And then the system froze after getting this:
```
[  857.148538] INFO: task kworker/u16:11:8004 blocked for more than 122 seconds.
[  857.148548]       Tainted: G           O       6.7.0-3-generic #3+scx5-Ubuntu
[  857.148550] "echo 0 > /proc/sys/kernel/hung_task_timeout_secs" disables this message.
[  857.148551] task:kworker/u16:11  state:D stack:0     pid:8004  tgid:8004  ppid:2      flags:0x00004000
[  857.148557] Workqueue: events_unbound bpf_map_free_deferred
[  857.148566] Call Trace:
[  857.148568]  <TASK>
[  857.148573]  __schedule+0x27c/0x6b0
[  857.148581]  schedule+0x33/0x110
[  857.148583]  schedule_timeout+0x157/0x170
[  857.148585]  wait_for_completion+0x88/0x150
[  857.148588]  __wait_rcu_gp+0x17e/0x190
[  857.148592]  bpf_struct_ops_map_free+0x52/0x90
[  857.148595]  ? __pfx_call_rcu+0x10/0x10
[  857.148599]  ? __pfx_call_rcu_tasks+0x10/0x10
[  857.148601]  ? __pfx_wakeme_after_rcu+0x10/0x10
[  857.148603]  bpf_map_free_deferred+0x55/0x90
[  857.148606]  process_one_work+0x16c/0x350
[  857.148610]  worker_thread+0x306/0x440
[  857.148613]  ? __pfx_worker_thread+0x10/0x10
[  857.148615]  kthread+0xef/0x120
[  857.148619]  ? __pfx_kthread+0x10/0x10
[  857.148621]  ret_from_fork+0x44/0x70
[  857.148625]  ? __pfx_kthread+0x10/0x10
[  857.148627]  ret_from_fork_asm+0x1b/0x30
[  857.148632]  </TASK>
```

arighi:
If you want to try the same kernel that I used: https://launchpad.net/~arighi/+archive/ubuntu/sched-ext-unstable (`ppa:arighi/sched-ext-unstable`)

(BTW, this is where I plan to upload new "proposed" Ubuntu sched-ext kernels, once everything looks fine I'll promote the kernel to `ppa:arighi/sched-ext`)

htejun:
I also saw `scx_userland` fail to init with `-ENOMEM` but it's not good that the system is freezing afterward. Will try to repro.

htejun:
@arighi I can't reproduce the problem you saw. How are you testing? If you can reproduce the problem && the test env is qemu, can you `dump-guest-memory -p vmcore.img` and share that somehow? Note that you'd need `CONFIG_FW_CFG_SYSFS` and `CONFIG_KEXEC` enabled for the dump to be useful.

arighi:
> @arighi I can't reproduce the problem you saw. How are you testing? If you can reproduce the problem && the test env is qemu, can you `dump-guest-memory -p vmcore.img` and share that somehow? Note that you'd need `CONFIG_FW_CFG_SYSFS` and `CONFIG_KEXEC` enabled for the dump to be useful.

This one seems to happen only on my laptop, unfortunately I wasn't able to reproduce it in qemu. Do you think your last commit can make any difference? I'm building a kernel with that extra patch right now.

htejun:
My last patch fixed `scx_rustland` getting stuck deep in the suspend path but probably won't fix the above issue unfortunately.

arighi:
ok I'll repeat the test with the new kernel and I'll try to collect more info if the problem happens again. Thanks!

arighi:
@htejun I'm running a kernel with this patch set applied (including the last commit), no problem at all on my laptop, I can correctly suspend/resume multiple times and scx_rustland is still functional. :tada: 

I'm rolling out this kernel to `ppa:arighi/sched-ext`, because everything seems stable, excluding https://github.com/sched-ext/scx/issues/70, but that's a separate issue, so this one looks good to me, thanks!

kernel/sched/ext.c
```diff
@@ -2904,22 +2920,96 @@ static void scx_cgroup_config_knobs(void) {}
...
+	    unlikely(scx_ops_enable_state() == SCX_OPS_DISABLING))
 		return false;
 	if (READ_ONCE(scx_switching_all))
 		return true;
 	return p->policy == SCHED_EXT;
 }
 
-static void scx_ops_fallback_enqueue(struct task_struct *p, u64 enq_flags)
+/**
+ * scx_ops_bypass - [Un]bypass scx_ops and guraantee forward progress
```
ByteLab-David: guarantee

kernel/sched/ext.c
```diff
@@ -4124,6 +4203,14 @@ void scx_bpf_kick_cpu(s32 cpu, u64 flags)
 		return;
 	}
 
+	/*
+	 * While bypassing for PM ops, IRQ handling may not be online which can
+	 * lead to irq_work_queue() malfunction such as infinite busy wait for
+	 * IRQ status update. Suppress kicking.
+	 */
```
ByteLab-David, line 4329: Can you add a comment mentioning that the cpu will be kicked when bypass is disabled to make it clear that aren't accidentally leaving the core idle?

Decave (approved):
I like it, this seems more elegant of a solution. Especially that we're now just skipping ops.enqueue() and ops.dispatch() instead of manually setting the ops. LGTM modulo two small nits

kernel/sched/ext.c
htejun: Kicking doesn't matter in that regard. What wakes up the CPUs is the enqueueing path calling `resched_curr()` if the CPU is in idle class, which is what we always depend on for not leaving CPUs idle when new tasks are put on them. The kicking in `scx_ops_bypass()` is to restart ticks as the scx scheduler may have disabled them which may confuse EEVDF if we end up switching to it.

PR #104, opened 2024-01-02
Allow dispatching from ops.select_cpu()

Decave:
A common pattern in schedulers is to find and reserve an idle core in
`ops.select_cpu()`, and to then use a task local storage map to specify
that the task should be enqueued in `SCX_DSQ_LOCAL` on the `ops.enqueue()`
path. At the same time, we also have a special `SCX_TASK_ENQ_LOCAL`
enqueue flag which is used by `scx_select_cpu_dfl()` to notify
`ops.enqueue()` that it may want to do a local enqueue.

Taking a step back, direct dispatch is something that should be
supported from the `ops.select_cpu()` path as well. At the end of the day,
it's on the same "enqueue path" as `ops.enqueue()`, and if we have the
contract be that doing a direct dispatch to `SCX_DSQ_LOCAL` will dispatch
the task to the local CPU of whatever is returned by `ops.select_cpu()`, it
should improve the ergonomics of `ops.select_cpu()` quite nicely. With that
in mind, let's just extend the API a bit to support direct dispatch from
`ops.select_cpu()`.

### DIFF
--- a/include/linux/sched/ext.h
+++ b/include/linux/sched/ext.h
@@ -184,6 +184,11 @@ struct sched_ext_ops {
 	 * If an idle CPU is returned, the CPU is kicked and will try to
 	 * dispatch. While an explicit custom mechanism can be added,
 	 * select_cpu() serves as the default way to wake up idle CPUs.
+	 *
+	 * @p may be dispatched directly by calling scx_bpf_dispatch(). If @p
+	 * is dispatched, the ops.enqueue() callback will be skipped. Finally,
+	 * if @p is dispatched to SCX_DSQ_LOCAL, it will be dispatched to the
+	 * local DSQ of whatever CPU is returned by this callback.
 	 */
 	s32 (*select_cpu)(struct task_struct *p, s32 prev_cpu, u64 wake_flags);
 
@@ -196,6 +201,9 @@ struct sched_ext_ops {
 	 * or enqueue on the BPF scheduler. If not directly dispatched, the bpf
 	 * scheduler owns @p and if it fails to dispatch @p, the task will
 	 * stall.
+	 *
+	 * If @p was dispatched from ops.select_cpu(), this callback is
+	 * skipped.
 	 */
 	void (*enqueue)(struct task_struct *p, u64 enq_flags);
 
@@ -597,7 +605,7 @@ struct scx_dispatch_q {
 enum scx_ent_flags {
 	SCX_TASK_QUEUED		= 1 << 0, /* on ext runqueue */
 	SCX_TASK_BAL_KEEP	= 1 << 1, /* balance decided to keep current */
-	SCX_TASK_ENQ_LOCAL	= 1 << 2, /* used by scx_select_cpu_dfl() to set SCX_ENQ_LOCAL */
+	SCX_TASK_DDSP_PRIQ	= 1 << 2, /* task should be enqueued on priq when directly dispatched */
 
 	SCX_TASK_OPS_PREPPED	= 1 << 8, /* prepared for BPF scheduler enable */
 	SCX_TASK_OPS_ENABLED	= 1 << 9, /* task has BPF scheduler enabled */
@@ -630,7 +638,7 @@ enum scx_kf_mask {
 	SCX_KF_CPU_RELEASE	= 1 << 2, /* ops.cpu_release() */
 	/* ops.dequeue (in REST) may be nested inside DISPATCH */
 	SCX_KF_DISPATCH		= 1 << 3, /* ops.dispatch() */
-	SCX_KF_ENQUEUE		= 1 << 4, /* ops.enqueue() */
+	SCX_KF_ENQUEUE		= 1 << 4, /* ops.enqueue() and ops.select_cpu() */
 	SCX_KF_REST		= 1 << 5, /* other rq-locked operations */
 
 	__SCX_KF_RQ_LOCKED	= SCX_KF_CPU_RELEASE | SCX_KF_DISPATCH |
@@ -684,6 +692,12 @@ struct sched_ext_entity {
 	 * recommended.
 	 */
 	u64			dsq_vtime;
+
+	/*
+	 * Used to track when a task has requested a direct dispatch from the
+	 * ops.select_cpu() path.
+	 */
+	u64			ddsq_id;
 
 	/*
 	 * If set, reject future sched_setscheduler(2) calls updating the policy
--- a/include/linux/sched/ext.h
+++ b/include/linux/sched/ext.h
@@ -639,11 +639,12 @@ enum scx_kf_mask {
 	/* ops.dequeue (in REST) may be nested inside DISPATCH */
 	SCX_KF_DISPATCH		= 1 << 3, /* ops.dispatch() */
 	SCX_KF_ENQUEUE		= 1 << 4, /* ops.enqueue() and ops.select_cpu() */
-	SCX_KF_REST		= 1 << 5, /* other rq-locked operations */
+	SCX_KF_SELECT_CPU	= 1 << 5, /* ops.select_cpu() */
+	SCX_KF_REST		= 1 << 6, /* other rq-locked operations */
 
 	__SCX_KF_RQ_LOCKED	= SCX_KF_CPU_RELEASE | SCX_KF_DISPATCH |
-				  SCX_KF_ENQUEUE | SCX_KF_REST,
-	__SCX_KF_TERMINAL	= SCX_KF_ENQUEUE | SCX_KF_REST,
+				  SCX_KF_ENQUEUE | SCX_KF_SELECT_CPU | SCX_KF_REST,
+	__SCX_KF_TERMINAL	= SCX_KF_ENQUEUE | SCX_KF_SELECT_CPU | SCX_KF_REST,
 };
 
 /*
--- a/kernel/sched/ext.c
+++ b/kernel/sched/ext.c
@@ -886,12 +886,8 @@ static void do_enqueue_task(struct rq *rq, struct task_struct *p, u64 enq_flags,
 	    (enq_flags & SCX_ENQ_LAST))
 		goto local;
 
-	if (!SCX_HAS_OP(enqueue)) {
-		if (enq_flags & SCX_ENQ_LOCAL)
-			goto local;
-		else
-			goto global;
-	}
+	if (!SCX_HAS_OP(enqueue))
+		goto global;
 
 	/* DSQ bypass didn't trigger, enqueue on the BPF scheduler */
 	qseq = rq->scx.ops_qseq++ << SCX_OPSS_QSEQ_SHIFT;
@@ -1793,7 +1789,7 @@ static void put_prev_task_scx(struct rq *rq, struct task_struct *p)
 		 * follow-up scheduling event.
 		 */
 		if (list_empty(&rq->scx.local_dsq.fifo))
-			do_enqueue_task(rq, p, SCX_ENQ_LAST | SCX_ENQ_LOCAL, -1);
+			do_enqueue_task(rq, p, SCX_ENQ_LAST, -1);
 		else
 			do_enqueue_task(rq, p, 0, -1);
 	}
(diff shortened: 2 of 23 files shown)


include/linux/sched/ext.h
```diff
@@ -112,6 +112,21 @@ enum scx_ops_flags {
...
+	 * default select CPU implementation from doing a direct dispatch, and
+	 * to instead have the ops.enqueue() callback invoked.
+	 *
+	 * If the default select CPU implementation finds an idle core, it will
+	 * invoke ops.enqueue() with the SCX_ENQ_LOCAL flag.
+	 *
+	 * If this flag is specified, ops.enqueue() must be defined, and
+	 * ops.select_cpu() may not be defined.
+	 */
+	SCX_OPS_ENQ_DFL_NO_DISPATCH= 1LLU << 3,
```
htejun: It's a word salad no matter what, so maybe it'd be better to name it more accurately? How about sth like `SCX_OPS_SELCPU_DFL_NO_DDSP` for "select_cpu_default no direct dispatch"?

htejun: I wonder whether we can skip the flag and require users to implement custom select_cpu if they want to override the behavior. e.g. we can provide `s32 scx_bpf_select_cpu_dfl(struct task_struct *p, s32 prev_cpu, u64 wake_flags, bool *end_direct` which is also used to implement `select_cpu_dfl`. Then, we can avoid adding the specialized flag while allowing users to achieve the behavior if desired.

kernel/sched/ext.c
```diff
@@ -3236,12 +3232,28 @@ static struct kthread_worker *scx_create_rt_helper(const char *name)
...
 }
 
+static int validate_ops(const struct sched_ext_ops *ops)
+{
+	/*
+	 * It doesn't make sense to specify the SCX_OPS_ENQ_LAST flag if the
+	 * ops.enqueue() callback isn't implemented.
+	 */
+	if ((ops->flags & SCX_OPS_ENQ_LAST) && !ops->enqueue)
+		return -EINVAL;
```
htejun, line 3244: Can we push this later so that we can fail with scx_ops_error() and report more details on why it's being failed?

include/linux/sched/ext.h
```diff
@@ -597,7 +605,7 @@ struct scx_dispatch_q {
 enum scx_ent_flags {
 	SCX_TASK_QUEUED		= 1 << 0, /* on ext runqueue */
 	SCX_TASK_BAL_KEEP	= 1 << 1, /* balance decided to keep current */
-	SCX_TASK_ENQ_LOCAL	= 1 << 2, /* used by scx_select_cpu_dfl() to set SCX_ENQ_LOCAL */
+	SCX_TASK_ENQ_DIRECT	= 1 << 2, /* task was directly dispatched */
```
htejun: Maybe "task is to be directly dispatched" is more intuitive?

kernel/sched/ext.c
```diff
@@ -814,16 +820,44 @@ static void direct_dispatch(struct task_struct *ddsp_task, struct task_struct *p
 		return;
 	}
 
+	WARN_ON_ONCE(p->scx.flags & SCX_TASK_ENQ_DIRECT);
+	WARN_ON_ONCE(p->scx.dsq_flags & SCX_TASK_DSQ_ON_PRIQ);
+	p->scx.flags |= SCX_TASK_ENQ_DIRECT;
+	if (enq_flags & SCX_ENQ_DSQ_PRIQ)
+		p->scx.dsq_flags |= SCX_TASK_DSQ_ON_PRIQ;
```
htejun: This probably is safe because while dsq_flags is synchronized the associated dsq's lock, the task here is not associated with a dsq and the field shouldn't be used. That said, I don't think it's a good idea to use this flag field like this. Let's just add another p->scx.flags field.

Another way to go about it is adding another `SCX_DSQ_FLAG_` bit. We have plenty of bits there and that would encode all the necessary information into `ddsq_id`.

kernel/sched/ext.c
```diff
@@ -814,16 +820,44 @@ static void direct_dispatch(struct task_struct *ddsp_task, struct task_struct *p
 		return;
 	}
 
+	WARN_ON_ONCE(p->scx.flags & SCX_TASK_ENQ_DIRECT);
+	WARN_ON_ONCE(p->scx.dsq_flags & SCX_TASK_DSQ_ON_PRIQ);
+	p->scx.flags |= SCX_TASK_ENQ_DIRECT;
```
htejun: Can't we use whether ddsq_id is INVALID? The code is initializing that value to INVALID but not depending on that anywhere.

kernel/sched/ext.c
```diff
@@ -814,16 +820,44 @@ static void direct_dispatch(struct task_struct *ddsp_task, struct task_struct *p
...
+}
+
+static void direct_dispatch(struct task_struct *p, u64 dsq_id, u64 enq_flags)
+{
+	struct scx_dispatch_q *dsq;
+
 	touch_core_sched_dispatch(task_rq(p), p);
 
+	WARN_ON_ONCE((dsq_id & SCX_DSQ_LOCAL_ON) == SCX_DSQ_LOCAL_ON);
+	WARN_ON_ONCE(!(p->scx.flags & SCX_TASK_ENQ_DIRECT));
```
htejun: This is definitely subjective but I feel this is too much WARN_ON_ONCE()'s. Both conditions are established recently by tightly coupled code, one just before the function is being called.

kernel/sched/ext.c
```diff
@@ -814,16 +820,44 @@ static void direct_dispatch(struct task_struct *ddsp_task, struct task_struct *p
...
+	WARN_ON_ONCE(p->scx.flags & SCX_TASK_ENQ_DIRECT);
+	WARN_ON_ONCE(p->scx.dsq_flags & SCX_TASK_DSQ_ON_PRIQ);
+	p->scx.flags |= SCX_TASK_ENQ_DIRECT;
+	if (enq_flags & SCX_ENQ_DSQ_PRIQ)
+		p->scx.dsq_flags |= SCX_TASK_DSQ_ON_PRIQ;
+
+	p->scx.ddsq_id = dsq_id;
+}
+
+static void direct_dispatch(struct task_struct *p, u64 dsq_id, u64 enq_flags)
```
htejun: Given that `mark_direct_dispatch()` directly modifies `p->ddsq_id`, it probably would make sense to make `direct_dispatch()` to consume and modify it directly too.

kernel/sched/ext.c
```diff
@@ -814,16 +820,44 @@ static void direct_dispatch(struct task_struct *ddsp_task, struct task_struct *p
...
-	dispatch_enqueue(dsq, p, enq_flags | SCX_ENQ_CLEAR_OPSS);
 
-	/*
-	 * Mark that dispatch already happened by spoiling direct_dispatch_task
-	 * with a non-NULL value which can never match a valid task pointer.
-	 */
-	__this_cpu_write(direct_dispatch_task, ERR_PTR(-ESRCH));
+	/* Don't use the priority queue if we fell back to the global DSQ */
+	if (unlikely(dsq->id != dsq_id)) {
+		WARN_ON_ONCE(dsq->id != SCX_DSQ_GLOBAL);
```
htejun: Again, this WARN_ON_ONCE() seems excessive. Besides, isn't this an already existing path where BPF scheduler can trigger `WARN_ON_ONCE` in `dispatch_enqueue()`? ie. Isn't the correct thing to do convert `dispatch_enqueue()` trigger `scx_ops_error()` instead of `WARN_ON_ONCE()` when `PRIQ` is specified for a builtin dsq?

The current code is really weird because the global dsq fallback path is already an `scx_ops_error()` path. The BPF scheduler is getting killed no matter what. But for that path, we are adding a work around to drop the `PRIQ` flag along with `WARN_ON_ONCE()` to check that the fallback is indeed `global_dsq` so that the error path doesn't trigger `WARN_ON_ONCE()` in `direct_dispatch()`. It's too elaborate for abort handling.

htejun (changes requested):
Overall LGTM. Some minor comments within.

include/linux/sched/ext.h
ByteLab-David: Sounds good, let's do that. Or, what if we just did `s32 scx_bpf_select_cpu_dfl(struct task_struct *p, s32 prev_cpu, u64 wake_flags)`?

Edit: nvm, that doesn't work because prev_cpu could have been idle.  Will apply your suggestion.

include/linux/sched/ext.h
Decave: Yeah that comment is weird. How about "task is recommended to be directly dispatched"? Given that it's really more of a recommendation than a directive.

kernel/sched/ext.c
ByteLab-David: Yeah, good idea

kernel/sched/ext.c
ByteLab-David: I don't feel strongly so will remove, but I generally like to err on the side of being liberal with code like this which checks invariants that a function is relying on. Though in the past I've worked on a kernel where such error checks are compiled out on "release" kernels, so I should probably adjust for Linux.

kernel/sched/ext.c
ByteLab-David: >Besides, isn't this an already existing path where BPF scheduler can trigger WARN_ON_ONCE in dispatch_enqueue()? ie. Isn't the correct thing to do convert dispatch_enqueue() trigger scx_ops_error() instead of WARN_ON_ONCE() when PRIQ is specified for a builtin dsq?

Yes, that's a great point. I'll make this change in a separate PR and will remove the excessive abort handling here.

include/linux/sched/ext.h
Decave: The more I think about it, the more I like this idea. It would allow us to get rid of both flags. You don't need SCX_ENQ_LOCAL if you can know from the BPF prog whether the cpu returned by scx_bpf_select_cpu_dfl() != prev_cpu.

kernel/sched/ext.c
ByteLab-David: I'll go with replacing the `SCX_TASK_ENQ_DIRECT` flag with a `SCX_TASK_DDSP_PRIQ` flag. It feels kind of weird to add a dsq flag given that it takes away a bit from non-built-in DSQ IDs for what's a temporary state.

include/linux/sched/ext.h
ByteLab-David: No longer applies now that we're using the ddsq_id.

ByteLab-David:
@htejun thanks for the thorough review -- please feel free to take another look when you have a moment.

kernel/sched/ext.c
htejun: We have 63 bits for builtin DSQ IDs when we have only 2 + nr_cpus builtin DSQs. So, we have room to spare there. But yeah either way is fine.

kernel/sched/ext.c
```diff
@@ -3234,6 +3290,10 @@ static int scx_ops_enable(struct sched_ext_ops *ops)
 	 */
 	cpus_read_lock();
 
+	ret = validate_ops(ops);
+	if (ret)
+		goto err_disable;
```
htejun: Would this end up calling `.exit` without calling `.init`?

tools/testing/selftests/scx/select_cpu_dfl.bpf.c
```diff
@@ -0,0 +1,38 @@
...
+#include <scx/common.bpf.h>
+
+char _license[] SEC("license") = "GPL";
+
+bool saw_local = false;
+
+void BPF_STRUCT_OPS(select_cpu_dfl_enqueue, struct task_struct *p,
+		    u64 enq_flags)
+{
+	if (enq_flags & SCX_ENQ_LOCAL)
```
htejun: This probably needs updating?

htejun (approved):
LGTM. Left a couple minor comments. Please feel free to merge. Longer term, I wonder whether it'd be possible to reduce the amount of boilerplating for test code.

ByteLab-David:
> LGTM. Left a couple minor comments. Please feel free to merge. Longer term, I wonder whether it'd be possible to reduce the amount of boilerplating for test code.

Yep, that's on my todo list. Will probably move it up because I'm finding it pretty convenient to write these testcases and plan to do more.

kernel/sched/ext.c
ByteLab-David: Yeah it would. I figured that made sense as an API for the contract to be, "If you've invoked the scheduler incorrectly you'll get a chance to hear about it in `.exit`, but I think it would be weird to also invoke `.init` if we know the configuration is invalid. Let me know if you disagree

htejun: I think it's a bit too surprising. init/exit being paired is a pretty natural assumption to make and we don't really lose anything by validating ops after calling `.init`, right?

ByteLab-David: It's subjective and I don't feel strongly so can make this change, but I think there's a chance someone could be confused if they see that `.init()` was called even if the scheduler was invalid (I'd expect static validation to happen before the scheduler is actually initialized by the user). On the other hand, the scheduler can still fail to load after `.init()` if e.g. `.prep_enable()` fails, so it's probably fine. Will apply and then merge

htejun: So, we can avoid both by not calling `.exit` if `.init` hasn't been called but I'm not sure the added code would be worth it. In the end, the only thing that really matters is that the user gets a diagnostic message which is understandable.

PR #105, opened 2024-01-04
Fix fallback

Decave:
We're currently checking whether a builtin DSQ is being used with priq
in scx_bpf_dispatch_vtime(). This neglects the fact that we could end up
falling back to scx_dsq_global if there's an error. If we error out with
SCX_ENQ_DSQ_PRIQ set in enqueue flags, we would trigger a warning in
dispatch_enqueue(). Let's instead just move the check to inside of
dispatch_enqueue().

### DIFF
--- a/kernel/sched/ext.c
+++ b/kernel/sched/ext.c
@@ -646,8 +646,20 @@ static void dispatch_enqueue(struct scx_dispatch_q *dsq, struct task_struct *p,
 		}
 	}
 
+	if (unlikely((dsq->id & SCX_DSQ_FLAG_BUILTIN) &&
+		     (enq_flags & SCX_ENQ_DSQ_PRIQ))) {
+		/*
+		 * SCX_DSQ_LOCAL and SCX_DSQ_GLOBAL DSQs always consume from
+		 * their FIFO queues. To avoid confusion and accidentally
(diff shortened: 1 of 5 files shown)


kernel/sched/ext.c
```diff
@@ -646,6 +646,19 @@ static void dispatch_enqueue(struct scx_dispatch_q *dsq, struct task_struct *p,
...
+		 * starving vtime-dispatched tasks by FIFO-dispatched tasks, we
+		 * disallow any internal DSQ from doing vtime ordering of
+		 * tasks.
+		 */
+		scx_ops_error("Cannot use vtime ordering for built-in DSQs");
+		enq_flags &= ~SCX_ENQ_DSQ_PRIQ;
+	}
+
 	if (enq_flags & SCX_ENQ_DSQ_PRIQ) {
 		WARN_ON_ONCE(dsq->id & SCX_DSQ_FLAG_BUILTIN);
```
htejun: Maybe drop this WARN_ON_ONCE?

htejun (approved):

kernel/sched/ext.c
ByteLab-David: Sorry, forgot to do that on my latest revision. Will drop and merge, thanks!

PR #115, opened 2024-01-09
Stash enq_flags when marking for direct dispatch

Decave:
We're currently not remembering enq flags during direct dispatch. Let's
record them in case someone wants to pass e.g. SCX_ENQ_PREEMPT from
ops.select_cpu().

Let's also reset ddsq_id and ddsq_enq_flags before calling
dispatch_enqueue() to ensure there's no races with the task being
consumed from another core.

### DIFF
--- a/include/linux/sched/ext.h
+++ b/include/linux/sched/ext.h
@@ -612,9 +612,8 @@ struct scx_dispatch_q {
 enum scx_ent_flags {
 	SCX_TASK_QUEUED		= 1 << 0, /* on ext runqueue */
 	SCX_TASK_BAL_KEEP	= 1 << 1, /* balance decided to keep current */
-	SCX_TASK_DDSP_PRIQ	= 1 << 2, /* task should be enqueued on priq when directly dispatched */
-	SCX_TASK_RESET_RUNNABLE_AT = 1 << 3, /* runnable_at should be reset */
-	SCX_TASK_DEQD_FOR_SLEEP	= 1 << 4, /* last dequeue was for SLEEP */
+	SCX_TASK_RESET_RUNNABLE_AT = 1 << 2, /* runnable_at should be reset */
+	SCX_TASK_DEQD_FOR_SLEEP	= 1 << 3, /* last dequeue was for SLEEP */
 
 	SCX_TASK_STATE_SHIFT	= 8,	  /* bit 8 and 9 are used to carry scx_task_state */
 	SCX_TASK_STATE_BITS	= 2,
@@ -689,7 +688,8 @@ struct sched_ext_entity {
 #ifdef CONFIG_SCHED_CORE
 	u64			core_sched_at;	/* see scx_prio_less() */
 #endif
-	u64			ddsq_id;
+	u64			ddsp_dsq_id;
+	u64			ddsp_enq_flags;
 
 	/* BPF scheduler modifiable fields */
 
--- a/kernel/sched/ext.c
+++ b/kernel/sched/ext.c
@@ -682,6 +682,13 @@ static void dispatch_enqueue(struct scx_dispatch_q *dsq, struct task_struct *p,
 	p->scx.dsq = dsq;
 
 	/*
+	 * scx.ddsq_id is only relevant on the direct dispatch path, but we
+	 * clear it here because the direct dispatch verdict may be overridden
+	 * on the enqueue path during e.g. bypass.
+	 */
+	p->scx.ddsq_id = SCX_DSQ_INVALID;
+
+	/*
 	 * We're transitioning out of QUEUEING or DISPATCHING. store_release to
 	 * match waiters' load_acquire.
 	 */
@@ -854,7 +861,6 @@ static void direct_dispatch(struct task_struct *p, u64 enq_flags)
 
 	dsq = find_dsq_for_dispatch(task_rq(p), p->scx.ddsq_id, p);
 	dispatch_enqueue(dsq, p, enq_flags | SCX_ENQ_CLEAR_OPSS);
-	p->scx.ddsq_id = SCX_DSQ_INVALID;
 }
 
 static bool test_rq_online(struct rq *rq)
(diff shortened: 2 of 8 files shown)


include/linux/sched/ext.h
```diff
@@ -690,6 +689,7 @@ struct sched_ext_entity {
 	u64			core_sched_at;	/* see scx_prio_less() */
 #endif
 	u64			ddsq_id;
+	u64			ddsq_enq_flags;
```
htejun: `ddsq_id` can be read as "direct dispatch q ID" but I'm not sure how to read `ddsq_enq_flags`. Maybe they should be `ddsp_dsq_id` (or `ddsp_qid`) and `ddsp_enq_flags`?

kernel/sched/ext.c
```diff
@@ -847,14 +845,11 @@ static void direct_dispatch(struct task_struct *p, u64 enq_flags)
 
 	touch_core_sched_dispatch(task_rq(p), p);
 
-	if (p->scx.flags & SCX_TASK_DDSP_PRIQ) {
-		enq_flags |= SCX_ENQ_DSQ_PRIQ;
-		p->scx.flags &= ~SCX_TASK_DDSP_PRIQ;
-	}
-
+	enq_flags |= p->scx.ddsq_enq_flags | SCX_ENQ_CLEAR_OPSS;
```
htejun: Can we update `scx_bpf_dispatch()` documentation explaining what happens to `enq_flags` when direct dispatching from `.select_cpu()`? This isn't immediately obvious.

tools/testing/selftests/scx/select_cpu_vtime.c
```diff
@@ -0,0 +1,50 @@
...
+		if (pids[i] == 0) {
+			sleep(1);
+			exit(0);
+		}
+	}
+
+	for (i = 0; i < NUM_CHILDREN; i++) {
+		SCX_EQ(waitpid(pids[i], &status, 0), pids[i]);
+		SCX_EQ(status, 0);
+	}
```
htejun: I have a hard time understanding what the test cases are testing. This looks like a basic vtime scheduler and we're forking and waiting a number of children. It'd be useful if each test case states what it aims to test and how it's doing that.

htejun (approved):
Generally LGTM. Minor comments.

kernel/sched/ext.c
ByteLab-David: Will do. FWIW, I think we should eventually consider updating scx_bpf_dispatch*() to not take enq_flags, and to instead take some flags that we define as part of the sched_ext API. Passing enq flags feels like it leaks implementation details of the core scheduler.

Edit: To clarify what I mean, enq flags includes flags that are passed to ops.enqueue() to signal things to the callback (e.g. SCX_ENQ_REENQ), and which don't really have semantic meaning in the context of dispatching. Right now, I think every scheduler just passes enq_flags through when calling scx_bpf_dispatch() (from ops.enqueue()) without knowing what's even in the flags. I think we should make it so the API is fully self contained. Flags passed to scx_bpf_dispatch() only mean something in the context of dispatching, not in the context of enqueuing.

tools/testing/selftests/scx/select_cpu_vtime.c
ByteLab-David: Sure, I'll document this more thoroughly. FWIW, the purposes of these testcases will become much more clear when I'm done with the test refactoring. The API for adding testcases will require the user to document what they're testing.

htejun: Yeah, maybe we drop the lower 32 bits which are shared with the core flags. That said, the shared flags we expose are `WAKEUP` and `HEAD`, both of which are well-defined and should be passed to enqueue.

kernel/sched/ext.c
htejun: Oh, left the comment in the wrong thread. Here it goes again: Yeah, we can drop the lower 32 bits of enq flags but the currently shared flags are `WAKEUP` and `HEAD`, both of which are well defined and should be passed to `ops.enqueue()`. We can redefine them so that we don't pass them through verbatim but I don't think it'd make whole lot of difference.

ByteLab-David: Oh sorry, let me try to be more clear. I think it's fine to have a 64 bit enq_flag field that shares the lower 32-bits with `ENQUEUE_WAKEUP` and `ENQUEUE_HEAD`. What I meant was that it feels kind of odd to also use those enq flags for dispatching. Most of the flags have no semantic meaning for dispatching. I think only `SCX_ENQ_HEAD` and `SCX_ENQ_PREEMPT` do, right? And `SCX_ENQ_PREEMPT` doesn't really make sense as an enq flag either. I was just proposing we add a new `enum scx_dsp_flags` type that included flags you can pass to dispatch.

ByteLab-David:
Addressed feedback and also added https://github.com/sched-ext/sched_ext/pull/115/commits/94cfcaced95b8dcc65b0c19e490a27e627a74112, which fixes direct dispatch from ops.select_cpu() for non-builtin DSQs

kernel/sched/ext.c
```diff
@@ -4101,13 +4098,18 @@ static void scx_dispatch_commit(struct task_struct *p, u64 dsq_id, u64 enq_flags
...
+ * and @p must match the task being enqueued. Also, %SCX_DSQ_LOCAL_ON can't be
+ * used to target the local DSQ of a CPU other than the enqueueing one. Use
+ * ops.select_cpu() to be on the target CPU in the first place.
  *
- * When called from ops.enqueue(), it's for direct dispatch and @p must match
- * the task being enqueued. Also, %SCX_DSQ_LOCAL_ON can't be used to target the
- * local DSQ of a CPU other than the enqueueing one. Use ops.select_cpu() to be
- * on the target CPU in the first place.
+ * When called from ops.select_cpu(), @enq_flags and @dsp_id are stored, and @p
+ * will be directly dispatched to the corresponding dispatch queue after
```
htejun, line 4116: It probably should also explain that the enq flags from the enqueue path are or'd before the task is dispatched as it actually has significance in how the task is queued on a dsq (the HEAD flag).

kernel/sched/ext.c
```diff
@@ -680,6 +680,7 @@ static void dispatch_enqueue(struct scx_dispatch_q *dsq, struct task_struct *p,
 	}
 	dsq->nr++;
 	p->scx.dsq = dsq;
+	p->scx.ddsq_id = SCX_DSQ_INVALID;
```
htejun: Please explain that we need ddsq_id cleared here because the stored direct dispatch verdict may be overridden in the enqueue path.

htejun (approved):

Issue #114, opened 2024-01-09
deadlock in scx_ops_bypass()?


arighi:
Just triggered the following while testing for a github workflow (using the latest 6.7 kernel from this repo).

I'll enable lockdep and repeat the test to see if we can get more info.

```
[   71.371592] rcu: INFO: rcu_preempt self-detected stall on CPU
[   71.371592] rcu: 	1-...!: (25906 ticks this GP) idle=70bc/0/0x1 softirq=2566/2566 fqs=0
[   71.371592] rcu: 	(t=26000 jiffies g=5665 q=1 ncpus=4)
[   71.371592] rcu: rcu_preempt kthread timer wakeup didn't happen for 25997 jiffies! g5665 f0x0 RCU_GP_WAIT_FQS(5) ->state=0x402
[   71.371592] rcu: 	Possible timer handling issue on cpu=0 timer-softirq=986
[   71.371592] rcu: rcu_preempt kthread starved for 26000 jiffies! g5665 f0x0 RCU_GP_WAIT_FQS(5) ->state=0x402 ->cpu=0
[   71.371592] rcu: 	Unless rcu_preempt kthread gets sufficient CPU time, OOM is now expected behavior.
[   71.371592] rcu: RCU grace-period kthread stack dump:
[   71.371592] task:rcu_preempt     state:I stack:14744 pid:17    tgid:17    ppid:2      flags:0x00004000
[   71.371592] Sched_ext: flatcg (enabled+all), task: runnable_at=-26000ms
[   71.371592] Call Trace:
[   71.371592]  <TASK>
[   71.371592]  ? schedule_timeout+0xc8/0x1b0
[   71.371592]  ? __pfx_process_timeout+0x10/0x10
[   71.371592]  ? rcu_gp_fqs_loop+0x145/0x830
[   71.371592]  ? rcu_gp_kthread+0x202/0x2e0
[   71.371592]  ? kthread+0xdd/0x110
[   71.371592]  ? __pfx_kthread+0x10/0x10
[   71.371592]  ? ret_from_fork+0x34/0x50
[   71.371592]  ? __pfx_kthread+0x10/0x10
[   71.371592]  ? ret_from_fork_asm+0x1b/0x30
[   71.371592]  </TASK>
[   85.635911] sched: RT throttling activated
[   85.635898] Sending NMI from CPU 1 to CPUs 0:
[   85.636892] NMI backtrace for cpu 0
[...]

PR #116, opened 2024-01-09
ci: add github workflow to test the sched-ext kernel

arighi:
Add a github action to test the sched-ext kernel with all the shipped schedulers.

The test uses a similar approach to the scx workflow [1], using virtme-ng to run each scheduler inside a sched-ext enabled kernel for a certain amount of time (30 sec) and checking for potential stall, oops or bug conditions.

In this case we can use `virtme-ng --build` to build a kernel with bare minimum support to run inside virtme-ng itself, instead of generating a fully featured kernel, to expedite the testing process.

The mandatory .config options required by sched-ext are stored in `.github/workflows/sched-ext.config` and they are passed to virtme-ng via the `--config` option.

The test itself is defined in `.github/workflows/run-schedulers`: the script looks for all the binaries in `tools/sched_ext/build/bin` and runs each one in a separate virtme-ng instance, to ensure that each run does not impact the others.

[1] https://github.com/sched-ext/scx/blob/main/.github/workflows/build-scheds.yml

### DIFF
--- a/.github/workflows/sched-ext.config
+++ b/.github/workflows/sched-ext.config
@@ -1,0 +1,34 @@
+# sched-ext mandatory options
+#
+CONFIG_BPF=y
+CONFIG_BPF_SYSCALL=y
+CONFIG_BPF_JIT=y
+CONFIG_DEBUG_INFO_BTF=y
+CONFIG_BPF_JIT_ALWAYS_ON=y
+CONFIG_BPF_JIT_DEFAULT_ON=y
+CONFIG_SCHED_CLASS_EXT=y
+
+# Enable scheduling debugging
+#
+CONFIG_SCHED_DEBUG=y
+
+# Enable extra scheduling features (for a better code coverage while testing
+# the schedulers)
+#
+CONFIG_SCHED_AUTOGROUP=y
+CONFIG_SCHED_CORE=y
+
+# Enable fully preemptible kernel for a better test coverage of the schedulers
+#
+# CONFIG_PREEMPT_NONE is not set
+# CONFIG_PREEMPT_VOLUNTARY is not set
+CONFIG_PREEMPT=y
+CONFIG_PREEMPT_COUNT=y
+CONFIG_PREEMPTION=y
+CONFIG_PREEMPT_DYNAMIC=y
+CONFIG_PREEMPT_RCU=y
+
+# Additional debugging information (useful to catch potential locking issues)
+#
+CONFIG_DEBUG_LOCKDEP=y
+CONFIG_DEBUG_ATOMIC_SLEEP=y
--- a/.github/workflows/run-schedulers
+++ b/.github/workflows/run-schedulers
@@ -1,0 +1,51 @@
+#!/bin/bash
+#
+# Run sched-ext scheduler for TIMEOUT seconds inside virtme-ng and catch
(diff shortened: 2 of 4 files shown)


.github/workflows/sched-ext.config
```diff
@@ -0,0 +1,31 @@
...
+CONFIG_BPF_JIT_ALWAYS_ON=y
+CONFIG_BPF_JIT_DEFAULT_ON=y
+CONFIG_SCHED_CLASS_EXT=y
+
+# Enable scheduling debugging
+#
+CONFIG_SCHED_DEBUG=y
+
+# Enable autogroup (better code coverage while testing the schedulers)
+CONFIG_SCHED_AUTOGROUP=y
```
htejun, line 18: We probably want `CONFIG_SCHED_CORE` too for coverage.

htejun (approved):

.github/workflows/sched-ext.config
ByteLab-David: I agree, but IMO it would be better served as a separate CI job that runs the tests with CONFIG_SCHED_CORE added, given that it arguably changes things quite a bit, and probably isn't what most people will be running with. This is fine for now though, we can split it later.

Decave (approved):
Fantastic, thanks so much for adding this!

.github/workflows/sched-ext.config
arighi: > I agree, but IMO it would be better served as a separate CI job that runs the tests with CONFIG_SCHED_CORE added, given that it arguably changes things quite a bit, and probably isn't what most people will be running with. This is fine for now though, we can split it later.

Too late... I added SCHED_CORE as well, but I can drop it again.

BTW, isn't there a boot option to disable  core scheduling? Like `nosmt` or similar? Or having the option enabled is enough to change the sched-ext core significantly?

ByteLab-David: >Too late... I added SCHED_CORE as well, but I can drop it again.

Oh that's fine, I meant we can split them out later. It's OK to include it now.

>BTW, isn't there a boot option to disable core scheduling? Like nosmt or similar? Or having the option enabled is enough to change the sched-ext core significantly?

Ahh yeah good point, maybe this is sufficient. I think we're guarding all of our core sched logic with sched_core_enabled() / sched_core_disabled(). So we'd only have to do a single build and then just boot with the different option which would save a good amount of time.

ByteLab-David:
I'll merge as soon as this next CI run finishes. Thanks again, @arighi

arighi:
> I'll merge as soon as this next CI run finishes. Thanks again, @arighi

Nope, I forgot to exclude `scx_layered` from the tests, otherwise it fails with the `No layer spec` error (so there's another CI run).

For this I was thinking, maybe we should define a very simple default layer spec, so that we can run layered with no option and include it as well in the tests.

htejun:
scx_layered already carries an example config within, so we can just use that if no spec is specified. I'll get to that later.

arighi:
And we just triggered a failure that I think it's related to #114 

At least the CI is doing its job.... :)

arighi:
Merged this, so that we can trigger more tests and see if we can reproduce the "int3" issue also using a "vanilla" sched-ext kernel.

PR #117, opened 2024-01-10
scx: Implement scx selftests framework

Decave:
We want to make it as easy as possible both to run tests, and to
    implement them. This means we ideally want a single test runner binary
    that can run the testcases, while also making it trivial to add a
    testcase without worrying about having to update the runner itself.
    
    To accomplish this, this patch adds a new declarative mechanism for
    defining scx tests by implementing a struct scx_test object. Tests can
    simply define such a struct, and then register it with the testrunner
    using a REGISTER_SCX_TEST macro. The build system will automatically
    compile the testcase and add machinery to have it be auto-registered
    into the runner binary. The runner binary then outputs test results in
    ktap [0] format so it can be consumed by CI systems.
    
    [0]: https://docs.kernel.org/dev-tools/ktap.html
    
    This PR implements the framework, and updates some of the testcases
    to use it. If the framework looks good, we can convert the remaining ones.


ByteLab-David:
Sorry for the commit being kinda messed up w/ some files not being properly `git mv`'d

htejun (approved):
This looks fantastic. Thanks for the great work.

PR #121, opened 2024-01-17
Make the pionter passing to .dispatch MAYBE_NULL

ThinkerYzu:
This patch marks the second argument of .dispatch with
PTR_MAYBE_NULL | PTR_TO_BTF_ID | PTR_TRUSTED in
bpf_scx_is_valid_access(). The verifier  will ensures the programs
always check if the argument is NULL before reading the pointed memory.

### DIFF
--- a/kernel/sched/ext.c
+++ b/kernel/sched/ext.c
@@ -3666,15 +3666,75 @@ const struct file_operations sched_ext_fops = {
 
 extern struct btf *btf_vmlinux;
 static const struct btf_type *task_struct_type;
+static u32 task_struct_type_id;
+
+/* Make the 2nd argument of .dispatch a pointer that can be NULL. */
+static bool promote_dispatch_2nd_arg(int off, int size,
+                                     enum bpf_access_type type,
+                                     const struct bpf_prog *prog,
+                                     struct bpf_insn_access_aux *info)
+{
+	const struct bpf_struct_ops *st_ops;
+	const struct btf_member *member;
+        const struct btf_type *t;
+        u32 btf_id, member_idx;
+	const char *mname;
+
+        /* btf_id should be the type id of struct sched_ext_ops */
+	btf_id = prog->aux->attach_btf_id;
+	st_ops = bpf_struct_ops_find(btf_id);
+	if (!st_ops)
+                return false;
+
+        /* BTF type of struct sched_ext_ops */
+        t = st_ops->type;
+
+	member_idx = prog->expected_attach_type;
+	if (member_idx >= btf_type_vlen(t))
+                return false;
+
+        /* Get the member name of this program.  For example, the
+         * member name of the dispatch program is "dispatch".
+         */
+	member = &btf_type_member(t)[member_idx];
+	mname = btf_name_by_offset(btf_vmlinux, member->name_off);
+
+        /* Chkeck if it is the 2nd argument of the function pointer at
+         * "dispatch" in struct sched_ext_ops. The arguments of
+         * struct_ops operators are placed in the context one after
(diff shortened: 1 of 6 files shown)


kernel/sched/ext.c
```diff
@@ -3672,6 +3672,36 @@ static bool bpf_scx_is_valid_access(int off, int size,
...
+                task_struct_btf_id = btf_find_by_name_kind(btf_vmlinux, "task_struct", BTF_KIND_STRUCT);
+                if (task_struct_btf_id < 0)
+                        return false;
+
+                info->reg_type = PTR_MAYBE_NULL | PTR_TO_BTF_ID | PTR_TRUSTED;
+                info->btf = btf_vmlinux;
+                info->btf_id = task_struct_btf_id;
+
+                return true;
+        }
```
htejun, line 3723: This looks great but can you please separate this out into its own function and add a bit of comment explaining what's happening?

htejun (changes requested):
Looks good to me. I left a comment asking for a trivial change.

kernel/sched/ext.c
```diff
@@ -3672,6 +3672,36 @@ static bool bpf_scx_is_valid_access(int off, int size,
...
+                return false;
+
+        t = st_ops->type;
+	member_idx = prog->expected_attach_type;
+	if (member_idx >= btf_type_vlen(t))
+                return false;
+
+	member = &btf_type_member(t)[member_idx];
+	mname = btf_name_by_offset(btf_vmlinux, member->name_off);
+        if (strcmp(mname, "dispatch") == 0 && off == sizeof(__u64) && type == BPF_READ) {
```
ByteLab-David: Could you please add a comment here explaining what we're checking here and why this corresponds to the `prev` arg of the callback?

kernel/sched/ext.c
```diff
@@ -3672,6 +3672,36 @@ static bool bpf_scx_is_valid_access(int off, int size,
...
+                info->reg_type = PTR_MAYBE_NULL | PTR_TO_BTF_ID | PTR_TRUSTED;
+                info->btf = btf_vmlinux;
+                info->btf_id = task_struct_btf_id;
+
+                return true;
+        }
+
 	if (off < 0 || off >= sizeof(__u64) * MAX_BPF_FUNC_ARGS)
 		return false;
 	if (type != BPF_READ)
```
ByteLab-David, line 3733: If you bring this check to the beginning of the function, you don't have to check `BPF_READ` above

tools/testing/selftests/scx/Makefile
```diff
@@ -175,6 +176,8 @@ testcase-targets := $(addsuffix .o,$(addprefix $(SCXOBJ_DIR)/,$(auto-test-target
 $(SCXOBJ_DIR)/runner.o: runner.c | $(SCXOBJ_DIR)
 	$(CC) $(CFLAGS) -c $< -o $@
 
+$(SCXOBJ_DIR)/maybe_null.o: $(INCLUDE_DIR)/maybe_null_fail.bpf.skel.h
+
```
ByteLab-David, line 180: This is fine -- in a follow-on change I'll update the recipe to check if there's a _fail variant of a prog and also take that as a conditional dependency

Decave (commented):
Thanks a lot for taking care of this. LGTM modulo nits

kernel/sched/ext.c
ThinkerYzu: Sure

kernel/sched/ext.c
ThinkerYzu: Agree

Decave (approved):

htejun (approved):

PR #122, opened 2024-01-18
scx: Replace /sys/kernel/debug/sched/ext with /sys/kernel/sched_ext

htejun:
The existing state interface in /sys/kernel/debug/sched/ext is not available if CONFIG_SCHED_DEBUG is not set. Replace it with a simpler /sys/kernel/sched_ext interface and add tools/sched_ext/scx_show_state.py to keep the same level of visibility.

This will need updates in documentation and scripts in the scx repo.

### DIFF
--- a/kernel/sched/ext.c
+++ b/kernel/sched/ext.c
@@ -3724,37 +3724,6 @@ err_disable:
 	return ret;
 }
 
-#ifdef CONFIG_SCHED_DEBUG
-static int scx_debug_show(struct seq_file *m, void *v)
-{
-	mutex_lock(&scx_ops_enable_mutex);
-	seq_printf(m, "%-30s: %s\n", "ops", scx_ops.name);
-	seq_printf(m, "%-30s: %ld\n", "enabled", scx_enabled());
-	seq_printf(m, "%-30s: %d\n", "switching_all",
-		   READ_ONCE(scx_switching_all));
-	seq_printf(m, "%-30s: %ld\n", "switched_all", scx_switched_all());
-	seq_printf(m, "%-30s: %s\n", "enable_state",
-		   scx_ops_enable_state_str[scx_ops_enable_state()]);
-	seq_printf(m, "%-30s: %d\n", "bypassing", scx_ops_bypassing());
-	seq_printf(m, "%-30s: %lu\n", "nr_rejected",
-		   atomic_long_read(&scx_nr_rejected));
-	mutex_unlock(&scx_ops_enable_mutex);
-	return 0;
-}
-
-static int scx_debug_open(struct inode *inode, struct file *file)
-{
-	return single_open(file, scx_debug_show, NULL);
-}
-
-const struct file_operations sched_ext_fops = {
-	.open		= scx_debug_open,
-	.read		= seq_read,
-	.llseek		= seq_lseek,
-	.release	= single_release,
-};
-#endif
-
 
 /********************************************************************************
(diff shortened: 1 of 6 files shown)


kernel/sched/ext.c
```diff
@@ -3390,6 +3474,19 @@ static int scx_ops_enable(struct sched_ext_ops *ops)
 		goto err;
 	}
 
+	scx_root_kobj = kzalloc(sizeof(*scx_root_kobj), GFP_KERNEL);
```
arighi, line 3477: We probably want to `kfree(src_root_kobj)` / `kobject_del()` in the other error paths at the end of `scx_ops_enable()`

arighi (approved):
Minor nit, everything else looks good to me. Thanks! This is very useful for distro kernels.

kernel/sched/ext.c
htejun: Other error paths are invoking `scx_ops_disable()` which eventually invokes `scx_ops_disable_workfn()` which `kobject_del()`s the kobj which in turn triggers `scx_kobj_release()`, so there shouldn't be any leaks. I'm gonna update the code a bit for clarity.

arighi: oh you're right, then nevermind. And yes, the `scx_root_kobj = NULL` close to the kfree / kobject_del makes the code more clear.

kernel/sched/ext.c
```diff
@@ -193,6 +193,10 @@ struct scx_dsp_ctx {
 
 static DEFINE_PER_CPU(struct scx_dsp_ctx, scx_dsp_ctx);
 
+/* /sys/kernel/sched_ext interface */
+struct kset *scx_kset;
+struct kobject *scx_root_kobj;
```
ByteLab-David: Can these be static?

tools/sched_ext/scx_show_state.py
```diff
@@ -0,0 +1,40 @@
...
+# Copyright (C) 2024 Meta Platforms, Inc. and affiliates.
+
+desc = """
+This is a drgn script to show the current sched_ext state.
+For more info on drgn, visit https://github.com/osandov/drgn.
+"""
+
+import sys
+
+import drgn
```
ByteLab-David: tiny nit: let's bring these imports together and alphabetize them?

kernel/sched/ext.c
```diff
@@ -3030,6 +3034,83 @@ static int scx_cgroup_init(void) { return 0; }
 static void scx_cgroup_config_knobs(void) {}
 #endif
 
+
+/********************************************************************************
+ * Sysfs interface and ops enable/disable.
+ */
+
+#define SCX_ATTR(_name)								\
```
ByteLab-David, line 3042: nit: to avoid ambiguity, could we please rename this to macro `SCX_KOBJ_ATTR`? It's dead clear from the context of where it's used what it's meant for, but ATTR seems like just too much of an overloaded term in sched contexts.

htejun: There is also `struct kobj_attribute`  which is a wrapper around `struct attribute`. The whole kobj thing is a mess - e.g. look at how `ATTRIBUTE_GROUPS()` macro is used in most places and how it introduces symbols named `SUBSYS_groups`. When it becomes too much, the usual way to organize is separating out kobject and sysfs related stuff into its own source file. We will get there later but for now, let's leave it be.

Decave (approved):

PR #125, opened 2024-01-22
Rebase onto 6.8-rc1

Decave:
Mostly a clean merge. Please take a look specifically at the following to make sure I didn't botch anything:

- `cgroup_add_file()` in kernel/cgroup/cgroup.c
- All of the changes in ext.c


htejun:
I'm unsure whether we want to be pulling in rc releases into the main branch for a couple reasons:
- `bpf/for-next` has always been our base when we needed to receive new bpf features. Note that `bpf/for-next` doesn't stay synced to `linus#master` and conflicts are possible when we pull in both `bpf/for-next` and `linus#master`.
- Early RCs tend to be pretty rough and pulling them in doesn't really give us any practical benefits.
- Tracking the latest stable release is likely more useful for most users including distros, which doesn't mean the devel branch is going to track stable kernels and no matter what we do there's going to be some manual work to cut and maintain kernel releases. The overall amount of work may be lower if we maintain -rcX kernel as separate releases rather than always pulling the into the main dev branch.

ByteLab-David:
>bpf/for-next has always been our base when we needed to receive new bpf features. Note that bpf/for-next doesn't stay synced to linus#master and conflicts are possible when we pull in both bpf/for-next and linus#master.

I was under the impression that we were going to be moving away from pulling bpf/for-next for the reason you mentioned? I assumed we'd just be tracking Linus' branch and doing merges as needed. Is the plan to keep merging bpf-next?

>Early RCs tend to be pretty rough and pulling them in doesn't really give us any practical benefits.

The practical benefit is that we're able to use and test our code against the latest "release" to uncover any issues; especially for bpf changes. I agree though that it's more convenient if we can be testing new features on top of a kernel we have confidence in, and that seems like the more reasonable thing to optimize for.

>Tracking the latest stable release is likely more useful for most users including distros, which doesn't mean the devel branch is going to track stable kernels and no matter what we do there's going to be some manual work to cut and maintain kernel releases. The overall amount of work may be lower if we maintain -rcX kernel as separate releases rather than always pulling the into the main dev branch.

Hmm, I think I'm a bit confused. If distros are tracking the latest stable release, then this should have no impact on them, correct? That said, I'm fine with not rebasing against rc kernels if that's your preference. I was originally going to cut a 6.8-rc1 release in https://github.com/sched-ext/scx-kernel-releases, but decided to do this first to save whomever cuts the next release the trouble of dealing with the merge conflicts.

htejun:
> I was under the impression that we were going to be moving away from pulling bpf/for-next for the reason you mentioned? I assumed we'd just be tracking Linus' branch and doing merges as needed. Is the plan to keep merging bpf-next?

I don't think we're moving way from basing on bpf/for-next necessarily. It just hasn't been necessary to pull in bpf/for-next for a while. The main reason we've been pulling bpf/for-next is because there often were something that we needed in bpf/for-next. Other than that, for the devel branch, we don't want to straggle too far behind linus#master while staying more or less stable.

There isn't a black-or-white answer to how close we want to track linus#master but there are costs associated with both tracking too close and falling behind too far.

Too close, we introduce instability to devel tree without actually getting much practical value out of it. If there are issues in linus#master, our dev tree experiencing the same issues is unlikely to be meaningful benefit either us or upstream. Another cost associated with following too close is higher cost of maintaining stable release kernel. As long as we're based on the latest release kernel, we can just keep pulling in changes. Once we move off to an rc, it isn't as simple anymore.

Lagging too far behind can bring surprises too late and if fall behind the latest release kernel, prepping kernels for distros becomes cumbersome too.

While not black and white, we can come up with rules of thumb. We don't want to fall behind fully released kernels and we don't want to pull in early rc's either. It likely makes the most sense to pull in later rc's which are close to release - say, rc5+ or so. At that point, both linus#master and -stable for the released kernel are a lot more stable, so the costs of both pulling in the new kernel and maintaining stable releases are a lot lower than earlier during the cycle.

Of course, if bpf/for-next has a feature that we want or some -rc release contains a necessary bug fix, we pull that in and eat the cost.
 
> > Tracking the latest stable release is likely more useful for most users including distros, which doesn't mean the devel branch is going to track stable kernels and no matter what we do there's going to be some manual work to cut and maintain kernel releases. The overall amount of work may be lower if we maintain -rcX kernel as separate releases rather than always pulling the into the main dev branch.
> 
> Hmm, I think I'm a bit confused. If distros are tracking the latest stable release, then this should have no impact on them, correct? That said, I'm fine with not rebasing against rc kernels if that's your preference. I was originally going to cut a 6.8-rc1 release in https://github.com/sched-ext/scx-kernel-releases, but decided to do this first to save whomever cuts the next release the trouble of dealing with the merge conflicts.

The cost for us is maintaining the `scx-kernel-releases` repo. If we want to maintain -rc releases too, I think it'd be better to keep them in `scx-kernel-releases`. As long as the fixes are in a few commits, forward carrying them with cherry-picks shouldn't be too bad and is easier to figure out than the other way around.

arighi:
On a similar topic, can we create a `sched_ext-ci` branch (or `sched_ext-stable`, or similar), so that the scx CI can use this branch, to make sure a regression in the kernel doesn't trigger failures in the schedulers?

(And obviously for the kernel CI, instead, we will still use the kernel from the latest push/pull request)

ByteLab-David:
Ack, will close this for now. We can rebase onto linus#master when 6.8 is released. For now I'll just do the rc releases on `scx-kernel-releases` as Peter Jung (@ptr1337) requested for CachyOS.

PR #130, opened 2024-01-28
More selftests

Decave:
Adds a couple of new selftest testcases:

- Loading a "maximal" scheduler that has every callback defined
- Loading schedulers repeatedly, concurrently in tight loops


ByteLab-David:
The test failure here is `scx_qmap` hitting the watchdog timeout. Given that these schedulers are running on shared github hosts in VMs, perhaps we should bump the timeout?

htejun (approved):

PR #132, opened 2024-01-29
Add fork field to struct scx_init_task_args

Decave:
Some schedulers, such as scx_nest, want to have different logic
depending on whether a task was running prior to the scheduler being
enabled. Let's add a flag that they can check to make it easy to find
out.

We also add a testcase and add a slight usability improvement to the
test runner.

### DIFF
--- a/include/linux/sched/ext.h
+++ b/include/linux/sched/ext.h
@@ -131,6 +131,11 @@ struct scx_init_task_args {
 	/* the cgroup the task is joining */
 	struct cgroup		*cgroup;
 #endif
+	/*
+	 * Set if ops.init_task() is being invoked on the fork path, as opposed
+	 * to the scheduler transition path.
+	 */
+	bool			fork;
 };
 
 /* argument container for ops.exit_task() */
--- a/kernel/sched/ext.c
+++ b/kernel/sched/ext.c
@@ -2367,7 +2367,7 @@ static void scx_set_task_state(struct task_struct *p, enum scx_task_state state)
 	p->scx.flags |= state << SCX_TASK_STATE_SHIFT;
 }
 
-static int scx_ops_init_task(struct task_struct *p, struct task_group *tg)
+static int scx_ops_init_task(struct task_struct *p, struct task_group *tg, bool fork)
 {
 	int ret;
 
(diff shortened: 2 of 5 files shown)


include/linux/sched/ext.h
```diff
@@ -131,6 +131,11 @@ struct scx_init_task_args {
 	/* the cgroup the task is joining */
 	struct cgroup		*cgroup;
 #endif
+	/*
+	 * Set if .init_task() is being invoked on the scheduler transition
+	 * path as opposed to the fork path.
+	 */
+	bool			transitioning;
```
htejun: I'm not sure about `transitioning` given that we don't use that verb in the interface elsewhere. Why not just `init` or `fork`? Probably `init` is better given that that's likely the special condition users would test for. If the concern is that `init_task_args.init` can be a bit confusing, maybe `sched_init`?

kernel/sched/ext.c
```diff
@@ -2335,6 +2335,11 @@ static struct cgroup *tg_cgrp(struct task_group *tg)
 
 #endif	/* CONFIG_EXT_GROUP_SCHED */
 
+#define SCX_INIT_TASK_ARGS(___tg, ___transitioning)	{	\
```
htejun: `SCX_INIT_TASK_ARGS_CGROUP()` macro is there because cgroup support is optional. Given that that's not the case for `init_task_args`, I don't see why we want to have this indirection.

include/linux/sched/ext.h
ByteLab-David: Hmm, I really don't like using `init` in an arg name, even if it's `sched_init`. Let's go with `fork`. Will respin shortly

kernel/sched/ext.c
ByteLab-David: I did it this way because it looks odd otherwise. We'd have to do this:

```
struct scx_init_task_args args = {
        SCX_INIT_TASK_ARGS_CGROUP(tg)
        .transitioning = transitioning,
};
```

Notice e.g. the lack of a comma, no `.cgroup =`. It's not a huge deal either way of course, but this seemed like it made the code more readable.

htejun: Maybe we can drop ',' from the cgroup macro?

ByteLab-David: That doesn't work, does it? Then you'd have an empty comma in the middle of the struct if EXT_GROUP_SCHED isn't defined.

ByteLab-David:
@htejun repushed. I didn't change the `SCX_INIT_TASK_ARGS()` thing because if you take the comma out of the macro it breaks the build. We can discuss further if you really dislike the extra layer of indirection.

htejun:
Yeah, please drop the macro. Thanks.

ByteLab-David:
> Yeah, please drop the macro. Thanks.

It's been dropped. FYI, this breaks auto-alignment for vim as well (though it only affects that specific section of code, not the whole file).

htejun (approved):

PR #136, opened 2024-02-04
scx: Only do CI runs for PRs

Decave:
We're basically always runnin two CI jobs: one for a remote push, and another for when a PR is opened. These are essentially measuring the same thing, so let's save CI bandwidth and just do a PR run. This will hopefully make things a bit less noisy as well.

### DIFF
--- a/.github/workflows/test-kernel.yml
+++ b/.github/workflows/test-kernel.yml
@@ -1,6 +1,6 @@
 name: test-kernel
 run-name: ${{ github.actor }} PR run
-on: [pull_request, push]
+on: [pull_request]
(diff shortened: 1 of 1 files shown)


htejun:
Looks fine to me but let's see what @arighi says. Maybe he had reasons to enable CI for both actions.

htejun (approved):

arighi (approved):
Initially I enabled both actions, so that we could do more tests with the CI itself, simply by pushing to any branch, but at this point I think the CI is stable enough and we can limit the runs only for PRs. Therefore, approved, thanks!

PR #138, opened 2024-02-07
scx: Implement SCX_KICK_IDLE

htejun:
Implement SCX_KICK_IDLE so that schedulers don't have to worry about causing too much churn with attempts to wake idle CPUs.

### DIFF
--- a/kernel/sched/ext.c
+++ b/kernel/sched/ext.c
@@ -4158,41 +4158,52 @@ static const struct sysrq_key_op sysrq_sched_ext_reset_op = {
 	.enable_mask	= SYSRQ_ENABLE_RTNICE,
 };
 
+static bool kick_one_cpu(s32 cpu, struct rq *this_rq, unsigned long *pseqs)
+{
+	struct rq *rq = cpu_rq(cpu);
+	struct scx_rq *this_scx = &this_rq->scx;
+	bool should_wait = false;
+	unsigned long flags;
+
+	raw_spin_rq_lock_irqsave(rq, flags);
+
+	/*
+	 * During CPU hotplug, a CPU may depend on kicking itself to make
+	 * forward progress. Allow kicking self regardless of online state.
+	 */
+	if (cpu_online(cpu) || cpu == cpu_of(this_rq)) {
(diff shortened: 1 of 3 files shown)


kernel/sched/ext.c
```diff
@@ -4158,49 +4166,109 @@ static const struct sysrq_key_op sysrq_sched_ext_reset_op = {
...
-	unsigned long *pseqs = this_cpu_ptr(scx_kick_cpus_pnt_seqs);
-	int this_cpu = cpu_of(this_rq);
-	int cpu;
+	lockdep_assert_rq_held(rq);
 
-	for_each_cpu(cpu, this_rq->scx.cpus_to_kick) {
-		struct rq *rq = cpu_rq(cpu);
-		unsigned long flags;
+	/*
+	 * We can skip idle kicking if @cpu is going to go through at least one
```
ByteLab-David: s/cpu/rq here and below

kernel/sched/ext.c
```diff
@@ -4158,49 +4166,109 @@ static const struct sysrq_key_op sysrq_sched_ext_reset_op = {
...
+	struct scx_rq *this_scx = &this_rq->scx;
+	bool should_wait = false;
+	unsigned long flags;
+
+	raw_spin_rq_lock_irqsave(rq, flags);
 
-		if (cpu_online(cpu) || cpu == this_cpu) {
-			if (cpumask_test_cpu(cpu, this_rq->scx.cpus_to_preempt) &&
-			    rq->curr->sched_class == &ext_sched_class)
+	if (cpu_online(cpu) || cpu == cpu_of(this_rq)) {
```
ByteLab-David, line 4199: So even though this functionally isn't a change from what was there before, would you mind please adding a comment explaining that a CPU can be offlined before IRQs are disabled to explain how it's possible that the CPU is offline but `cpu == cpu_of(this_rq)`?

Decave (approved):

kernel/sched/ext.c
htejun: It's just that CPU on/offlining is a multi-stage operation and cpu hotplug ops themselves depend on scheduling working. I'm not sure how much detail we want to explain here. I'll mention it briefly.

PR #139, opened 2024-02-07
scx: Actually allocate rq->scx.cpus_to_kick_if_idle

htejun:
To avoid crashing the kernel when CONFIG_CPUMASK_OFFSTACK is set.


Decave (approved):
Sorry for missing this in review

arighi (commented):
Looks good, thanks for fixing this!

Issue #142, opened 2024-02-14
memory leak in `scx_ops_enable`


xingyiz:
Syzkaller found a memory leak in `scx_ops_enable`. Looking at the code, it seems that the allocation is being allocated and freed properly so we're not exactly sure what causes the unreferenced object, but this issue has popped up quite frequently from Syzkaller that we thought you might want to take a look. 

Commit version: https://github.com/sched-ext/sched_ext/commit/65c265174d11b9792840ff18f166dd8c4b89d9b0
Config file: 
[config.txt](https://github.com/sched-ext/sched_ext/files/14279261/config.txt)

Bug report:
```
BUG: memory leak
unreferenced object 0xffff88800b935600 (size 64):
  comm "scx_serialise", pid 736, jiffies 4295779783 (age 80.593s)
  hex dump (first 32 bytes):
    c0 07 68 83 ff ff ff ff 08 56 93 0b 80 88 ff ff  ..h......V......
    08 56 93 0b 80 88 ff ff 00 00 00 00 00 00 00 00  .V..............
  backtrace:
    [<00000000cf39be81>] kmemleak_alloc_recursive include/linux/kmemleak.h:42 [inline]
    [<00000000cf39be81>] slab_post_alloc_hook mm/slab.h:766 [inline]
    [<00000000cf39be81>] slab_alloc_node mm/slub.c:3478 [inline]
    [<00000000cf39be81>] __kmem_cache_alloc_node+0x1f1/0x310 mm/slub.c:3517
    [<00000000a2522a89>] kmalloc_trace+0x26/0xc0 mm/slab_common.c:1098
    [<00000000ffb5fd6b>] kmalloc include/linux/slab.h:600 [inline]
    [<00000000ffb5fd6b>] kzalloc include/linux/slab.h:721 [inline]
    [<00000000ffb5fd6b>] scx_ops_enable+0x234/0x1610 kernel/sched/ext.c:3632
    [<00000000975486ca>] bpf_struct_ops_link_create+0x21e/0x290 kernel/bpf/bpf_struct_ops.c:914
    [<00000000da36584a>] link_create kernel/bpf/syscall.c:4938 [inline]
    [<00000000da36584a>] __sys_bpf+0x3684/0x4970 kernel/bpf/syscall.c:5453
    [<00000000418ac434>] __do_sys_bpf kernel/bpf/syscall.c:5487 [inline]
[...]

PR #143, opened 2024-02-14
scx: Avoid memory leak of scx_root_kobj

Decave:
We're making two mistakes with respect to object lifecycle for scx_root_kobj:

- We're not calling kobject_put() if kobject_init_and_add() returns an
   error code. According to the API for that function, we should be.
    
- We weren't kobject_putt()'ing scx_root_kobj on the
   scx_ops_disable_workfn() path. We were calling kobject_del(), but this
   doesn't actually call the release callback under the hood. It just
   unlinks the kobject from the hierarchy.

Address both of these. With these changes, we no longer trigger the memory leak detected by Syzkaller in
https://github.com/sched-ext/sched_ext/issues/142. Without this patch, we see the following with memleak enabled:

unreferenced object 0xffff8dd78a99f000 (size 64):
  comm "runner", pid 2383, jiffies 4294697813 (age 128.878s)
  hex dump (first 32 bytes):
    78 17 5f be ff ff ff ff 08 f0 99 8a d7 8d ff ff  x._.............
    08 f0 99 8a d7 8d ff ff 00 00 00 00 00 00 00 00  ................
  backtrace:
    [<0000000034bda486>] __kmem_cache_alloc_node+0x109/0x1f0
    [<00000000a91612ee>] kmalloc_trace+0x29/0xa0
    [<00000000bfe2a4b7>] bpf_scx_reg+0x135/0xd80
    [<000000000e3a5728>] bpf_struct_ops_link_create+0xd7/0x130
    [<00000000b14e1608>] __sys_bpf+0x2a9/0x530
    [<000000003140300e>] __x64_sys_bpf+0x17/0x20
    [<00000000fa410384>] do_syscall_64+0x4d/0xf0
    [<000000008cbc0d51>] entry_SYSCALL_64_after_hwframe+0x6f/0x77

With it, we see no such entries.

Fixes: e7a7781b485d4 ("scx: Add /sys/kernel/sched_ext interface")

### DIFF
--- a/kernel/sched/ext.c
+++ b/kernel/sched/ext.c
@@ -3395,7 +3395,14 @@ static void scx_ops_disable_workfn(struct kthread_work *work)
 		SCX_CALL_OP(SCX_KF_UNLOCKED, exit, ei);
 
 	cancel_delayed_work_sync(&scx_watchdog_work);
+	/*
+	 * Delete the kobject from the hierarchy eagerly in addition to just
+	 * dropping a reference. Otherwise, if the object is deleted
+	 * asynchronously, sysfs could observe an object of the same name still
+	 * in the hierarchy when another scheduler is loaded.
+	 */
 	kobject_del(scx_root_kobj);
+	kobject_put(scx_root_kobj);
 	scx_root_kobj = NULL;
 
 	memset(&scx_ops, 0, sizeof(scx_ops));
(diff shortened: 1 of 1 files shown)


kernel/sched/ext.c
```diff
@@ -3865,10 +3872,8 @@ static int scx_ops_enable(struct sched_ext_ops *ops)
 	return 0;
 
 err:
-	if (scx_root_kobj) {
-		kfree(scx_root_kobj);
-		scx_root_kobj = NULL;
-	}
+	kobject_put(scx_root_kobj);
```
ByteLab-David, line 3877: We could potentially still hit the issue of the sysfs node still being present if the root kobj is deleted asynchronously here, but in practice I'm not sure that matters as it's only possibly going to be an issue if `CONFIG_DEBUG_KOBJECT_RELEASE` is defined. Maybe we should add another goto label that we call from  the alloc_exit_info() path? Hmm, yeah, let me do that.

htejun (approved):

PR #150, opened 2024-02-28
scx: Improve error reporting after an early scx_bpf_error()

htejun:
If scx_bpf_error() is triggered while a BPF scheduler is being loaded, depending on the point of failure, it can make ops_state transition fail and thus make loading fail with -EBUSY, which isn't very descriptive.

There was already a workaround for a specific case where ops.init() triggers scx_bpf_error() which works by just reporting success so that the usual exit path report the error with all the details.

There's no reason to restrict this to ops.init() failure. These early scx_bpf_error()s always make the ops_state transitions fail and we can always finish loading with success and let the exit path report the faliure.

### DIFF
--- a/kernel/sched/ext.c
+++ b/kernel/sched/ext.c
@@ -3693,17 +3693,6 @@ static int scx_ops_enable(struct sched_ext_ops *ops)
 			ret = ops_sanitize_err("init", ret);
 			goto err_disable_unlock_cpus;
 		}
-
-		/*
-		 * Exit early if ops.init() triggered scx_bpf_error(). Not
-		 * strictly necessary as we'll fail transitioning into ENABLING
-		 * later but that'd be after calling ops.init_task() on all
-		 * tasks and with -EBUSY which isn't very intuitive. Let's exit
-		 * early with success so that the condition is notified through
-		 * ops.exit() like other scx_bpf_error() invocations.
-		 */
-		if (atomic_read(&scx_exit_kind) != SCX_EXIT_NONE)
-			goto err_disable_unlock_cpus;
 	}
 
 	for (i = SCX_OPI_CPU_HOTPLUG_BEGIN; i < SCX_OPI_CPU_HOTPLUG_END; i++)
@@ -3827,11 +3816,17 @@ static int scx_ops_enable(struct sched_ext_ops *ops)
 	/*
 	 * From here on, the disable path must assume that tasks have ops
 	 * enabled and need to be recovered.
+	 *
+	 * Transition to ENABLING fails iff the BPF scheduler has already
+	 * triggered scx_bpf_error(). Returning an error code here would lose
+	 * the recorded error information. Exit indicating success so that the
+	 * error is notified through ops.exit() with all the details.
 	 */
 	if (!scx_ops_tryset_enable_state(SCX_OPS_ENABLING, SCX_OPS_PREPPING)) {
 		preempt_enable();
 		spin_unlock_irq(&scx_tasks_lock);
-		ret = -EBUSY;
+		WARN_ON_ONCE(atomic_read(&scx_exit_kind) == SCX_EXIT_NONE);
+		ret = 0;
 		goto err_disable_unlock_all;
 	}
 
@@ -3865,8 +3860,10 @@ static int scx_ops_enable(struct sched_ext_ops *ops)
 	cpus_read_unlock();
 	percpu_up_write(&scx_fork_rwsem);
 
+	/* see above ENABLING transition for the explanation on exiting with 0 */
 	if (!scx_ops_tryset_enable_state(SCX_OPS_ENABLED, SCX_OPS_ENABLING)) {
-		ret = -EBUSY;
+		WARN_ON_ONCE(atomic_read(&scx_exit_kind) == SCX_EXIT_NONE);
+		ret = 0;
 		goto err_disable;
 	}
 


kernel/sched/ext.c
```diff
@@ -3827,11 +3816,17 @@ static int scx_ops_enable(struct sched_ext_ops *ops)
 	/*
 	 * From here on, the disable path must assume that tasks have ops
 	 * enabled and need to be recovered.
+	 *
+	 * Transition to ENABLING fails iff the BPF scheduler has already
```
jordalgo, line 3820:
```suggestion
	 * Transition to ENABLING fails if the BPF scheduler has already
```

jordalgo (approved):
comment nit but LGTM otherwise

kernel/sched/ext.c
```diff
@@ -3827,11 +3816,17 @@ static int scx_ops_enable(struct sched_ext_ops *ops)
...
+	 * triggered scx_bpf_error(). Returning an error code here would lose
+	 * the recorded error information. Exit indicating success so that the
```
ByteLab-David, line 3822: Sorry, I'm a bit confused -- why do we need to return 0 in order for the exit path to print the full failure? Shouldn't the kthread run regardless of whether we successfully register the struct_ops? That's why we have the `scx_ops_tryset_enable_state` check, right? Because we can only change the enable state from `scx_ops_disable_workfn()`.

kernel/sched/ext.c
```diff
@@ -3693,17 +3693,6 @@ static int scx_ops_enable(struct sched_ext_ops *ops)
...
-		/*
-		 * Exit early if ops.init() triggered scx_bpf_error(). Not
-		 * strictly necessary as we'll fail transitioning into ENABLING
-		 * later but that'd be after calling ops.init_task() on all
-		 * tasks and with -EBUSY which isn't very intuitive. Let's exit
-		 * early with success so that the condition is notified through
-		 * ops.exit() like other scx_bpf_error() invocations.
-		 */
-		if (atomic_read(&scx_exit_kind) != SCX_EXIT_NONE)
-			goto err_disable_unlock_cpus;
```
ByteLab-David, line 3706: IIRC I added this code, but I'm not remembering why we need to return with 0 to get exit information. If we do, do we need to handle possible failures below as well? E.g. after calling `validate_ops`, `__alloc_percpu`, etc. I think we definitely do for `validate_ops()` because we call `scx_ops_error()` in it.

kernel/sched/ext.c
htejun: That's "if and only if".

kernel/sched/ext.c
htejun: Well, if we fail from init path, the userspace wouldn't run the usual exit path. It would just see the error code returned from load and that's all the context that will have. If we succeed init knowing that ops_error is pending, then the userspace will run the normal exit path with the richer error reporting.

kernel/sched/ext.c
htejun: You didn't and no. This only applies to the cases where ops_error triggered. Not synchronous in-line init failures.

kernel/sched/ext.c
ByteLab-David: I see, so we want the user space exit info output in addition to the `pr_err()` / `printk()` output from the kthread. Fair enough. Would you mind please adding that to the comment before merging?

kernel/sched/ext.c
ByteLab-David: Ah that's right, I think I was the one who noticed it was an issue and you fixed it.

kernel/sched/ext.c
htejun: That's what the comment already says, no?

ByteLab-David: I meant adding, "in addition to the error message outputted to the console/dmesg by the scx_ops_disable_work kthread." I think it's probably useful to explain given that the error output is also emitted there, no?

htejun: I don't see much value in adding that and generally don't like to go into too much detail of a separate piece of code unless they're intimately tied. How the exit path reports error conditions is independent from the init path. From init path's POV, the only choice to make is whether to report scx_ops_error() as a part of init failure or defer it to the normal exit path which has a richer interface. How the exit path exactly works isn't a detail that the init path should explain.

ByteLab-David: It's a comment, so I can't really say that I care too much one way or the other, but I do think it's confusing as written now. "Returning an error code here would lose the recorded error information" isn't really true. It prevents us from plumbing the error information through ops.exit(), but the error information is retained and printed on the exit path here: https://github.com/sched-ext/sched_ext/blob/sched_ext/kernel/sched/ext.c#L3388-L3397. Anyways, like I said, I don't care strongly one way or the other. I'll just merge it as is.

kernel/sched/ext.c
jordalgo: TIL! But is this common in kernel comments / outside the math world?

htejun: I think it's common enough.
```
htejun@slm ~/o/work (htejun)> git grep ' iff ' | wc -l
437
```

ByteLab-David: At this point I don't even see 'iff'. My brain just interpolates 'if and only if' without me even realizing it

jordalgo: I don't trust that people aren't just misspelling 😆  . Ok, ok, I withdraw my objection to this shorthand.

PR #152, opened 2024-03-05
scx: Ignore WF_EXEC

htejun:
sched_exec() calls select_task_rq() and then do a rq migration if the returned cpu is different from current. This calls ops.select_cpu() outside the normal enqueue path confusing the direct dispatch mechanism - e.g. rq->scx.ddsp_dsq_id can be asserted when it shouldn't be.

I tried to make ops.select_cpu() work in this path but the mechanism doesn't make whole lot of sense for scx and it makes the semantics around ops.select_cpu() unnecessarily complicated for rare corner cases. If we want this behavior, I think the right thing to do is adding a separate method which can preempt self to trigger the normal scheduling path for the task.

### DIFF
--- a/kernel/sched/ext.c
+++ b/kernel/sched/ext.c
@@ -2154,6 +2154,19 @@ __bpf_kfunc_end_defs();
 
 static int select_task_rq_scx(struct task_struct *p, int prev_cpu, int wake_flags)
 {
+	/*
+	 * sched_exec() calls with %WF_EXEC when @p is about to exec(2) as it
+	 * can be a good migration opportunity with low cache and memory
+	 * footprint. Returning a CPU different than @prev_cpu triggers
+	 * immediate rq migration. However, for SCX, as the current rq
+	 * association doesn't dictate where the task is going to run, this
+	 * doesn't fit well. If necessary, we can later add a dedicated method
+	 * which can decide to preempt self to force it through the regular
+	 * scheduling path.
+	 */
+	if (unlikely(wake_flags & WF_EXEC))
+		return prev_cpu;
+
 	if (SCX_HAS_OP(select_cpu)) {
 		s32 cpu;
 		struct task_struct **ddsp_taskp;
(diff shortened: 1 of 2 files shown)


kernel/sched/ext.c
```diff
@@ -2154,6 +2154,19 @@ __bpf_kfunc_end_defs();
...
 {
+	/*
+	 * sched_exec() calls with %WF_EXEC when @p is about to exec(2) as it
+	 * can be a good migration opportunity with low cache and memory
+	 * footprint. Returning a CPU different than @prev_cpu triggers
+	 * immediate rq migration. However, for SCX, as the current rq
+	 * association doesn't dictate where the task is going to run, this
+	 * doesn't fit well. If necessary, we can later add a dedicated method
+	 * which can decide to preempt self to force it through the regular
+	 * scheduling path.
```
ByteLab-David, line 2165: Hmm, so this would just require us to implement `wakeup_preempt_scx()`, right? And the idea is that we'd add a new ops callback that indicates whether we should invoke `resched_curr()`? Makes sense.

Decave (approved):

dschatzberg:
The code looks fine to me. I don't fully understand the challenge here though.

> but the mechanism doesn't make whole lot of sense for scx

What do you mean by this?

kernel/sched/ext.c
htejun: I'm not sure it'd be `wakeup_preempt_scx()`. We probably want to add another method in the above if block, say `ops.on_exec()` or something. The scheduler implementation can then call e.g. `scx_bpf_kick(SCX_KICK_PREEMPT)` on self if it wants a full scheduling on itself at that point. Overall, tho, I'm not sure how big an optimization opportunity this is. Especially for schedules which would use cgroups and other attributes for placement, the decision point would be when e.g. `systemd`'s new child enters a new cgroup, not when it `exec`s.

htejun:
> The code looks fine to me. I don't fully understand the challenge here though.
> 
> > but the mechanism doesn't make whole lot of sense for scx
> 
> What do you mean by this?

It triggers the rq migration path in scheduler core which is implemented by scheduling a highpri RT thread on the local CPU which then changes the CPU of the victim thread that it preempted, which doesn't really fit with what SCX does with rq association - e.g. core code can do all that but scx can still hot-migrate the task when the victim task actually starts running, so the whole thing is rather pointless.

PR #156, opened 2024-03-06
scx: Replace scx_bpf_switch_all() with SCX_OPS_SWITCH_PARTIAL

htejun:
Now that ops fields can be modified before loading, replace scx_bpf_switch_all() with SCX_OPS_SWITCH_PARTIAL.


tools/sched_ext/include/scx/compat.h
```diff
@@ -0,0 +1,83 @@
...
+#include <errno.h>
+
+struct btf *__COMPAT_vmlinux_btf __attribute__((weak));
+
+static inline void __COMPAT_load_vmlinux_btf(void)
+{
+	if (!__COMPAT_vmlinux_btf) {
+		__COMPAT_vmlinux_btf = btf__load_vmlinux_btf();
+		SCX_BUG_ON(!__COMPAT_vmlinux_btf, "btf__load_vmlinux_btf() (%s)",
+			   strerror(errno));
```
ByteLab-David: `SCX_BUG_ON()` already prints `errno`

tools/sched_ext/include/scx/compat.bpf.h
```diff
@@ -7,12 +7,30 @@
...
 	else
-		scx_bpf_kick_cpu(cpu, 0);
+		return 0;
+}
+
+/*
+ * %SCX_KICK_IDLE is a later addition. To support both before and after, use
+ * %__COMPAT_SCX_KICK_IDLE which becomes 0 on kernels which don't support it.
+ */
+#define __COMPAT_SCX_KICK_IDLE __COMPAT_SCX_KICK_IDLE()
```
ByteLab-David: Maybe something like this would be more extensible so we don't have to manually add a function every time we want to check for an enum existing?

```
#define __COMPAT_ENUM_OR_ZERO(___type, __value)  \
({  \
        ___type ___ret = 0;  \
        if (bpf_core_enum_value_exists((___type), (___value)) \
                 ___ret = ___value; \  
        ___ret; \
)}
```

Decave (approved):

PR #158, opened 2024-03-07
scx: Make exit debug dump buffer resizable

htejun:
Schedulers can now request custom debug dump buffer size by setting `sched_ext_ops.exit_dump_len`. This will help debugging on bigger and busier systems.

### DIFF
--- a/tools/sched_ext/include/scx/compat.h
+++ b/tools/sched_ext/include/scx/compat.h
@@ -116,7 +116,8 @@ static inline bool __COMPAT_struct_has_field(const char *type, const char *field
  *   it, the value is ignored and a warning is triggered if the value is
  *   requested to be non-zero.
  */
-#define SCX_OPS_LOAD(__skel, __ops_name, __scx_name) ({				\
+#define SCX_OPS_LOAD(__skel, __ops_name, __scx_name, __uei_name) ({		\
+	UEI_SET_SIZE(__skel, __ops_name, __uei_name);				\
 	if (__COMPAT_struct_has_field("sched_ext_ops", "exit_dump_len")) {	\
 		bpf_map__set_autocreate((__skel)->maps.__ops_name, true);	\
 		bpf_map__set_autocreate((__skel)->maps.__ops_name##___no_exit_dump_len, false); \
--- a/tools/sched_ext/include/scx/compat.h
+++ b/tools/sched_ext/include/scx/compat.h
@@ -69,6 +69,34 @@ static inline bool __COMPAT_read_enum(const char *type, const char *name, u64 *v
 	__val;									\
 })
 
+static inline bool __COMPAT_struct_has_field(const char *type, const char *field)
+{
+	const struct btf_type *t;
+	const struct btf_member *m;
+	const char *n;
+	s32 tid;
+	int i;
+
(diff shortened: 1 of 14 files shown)


ByteLab-David:
Is there perhaps a middle ground between having a completely static dump size, and having to require the user to pass a maximum buffer size? For example, having a maximum amount of memory per core for the system, so that we do scale with size (or something to that effect)?

My worry with this is that it's a difficult knob for a user to know how to set correctly, and it's memory that we actually are allocating when the scheduler is loaded so going with a large buffer would be wasteful.

htejun:
We can always make the default amount adaptive but I'm not sure what that the criteria should be. The problem is that the amount of dump generated is highly dependent on what's running on the system, which has some correlation with machine size but that's not very strong. Also, the amount allocated for dump buffering is likely completely wasted in most cases, especially for users, and this just isn't something that end users would really care about. The dump code should probably do a better job of printing out more pertinent information at the top tho.

ByteLab-David:
> We can always make the default amount adaptive but I'm not sure what that the criteria should be.

Hmm, it seems to me like we'd be better at judging that than a user though, no? E.g. being able to scale it by num cores, max # of tasks to print per rq, etc. Though I do agree that it would be sad if whatever we chose was less than someone was willing to give, and ended up losing output because of it.

>The problem is that the amount of dump generated is highly dependent on what's running on the system, which has some correlation with machine size but that's not very strong.

Makes sense. E.g. some pathological overcommit cases could need a lot more buffer than another workload on the same host.

>Also, the amount allocated for dump buffering is likely completely wasted in most cases, especially for users, and this just isn't something that end users would really care about. The dump code should probably do a better job of printing out more pertinent information at the top tho.

Agreed that it's not a knob that most people will care about. It just feels like a very leaky API though. But also, it's not very uncommon to allow people to configure log buffer sizes.

Thanks for talking through it.

Byte-Lab (approved):
Looks reasonable, thanks for adding all of the BC macros as well. One request: would you mind please adding a testcase to the sefltests dir that verifies loading a scheduler with a large and/or small buffer? Just so that we exercise the custom path a bit.

tools/sched_ext/include/scx/compat.h
```diff
@@ -78,4 +106,38 @@ static inline bool __COMPAT_read_enum(const char *type, const char *name, u64 *v
...
+ *
+ * - sched_ext_ops.exit_dump_len was added later. On kernels which don't support
+ *   it, the value is ignored and a warning is triggered if the value is
+ *   requested to be non-zero.
+ */
+#define SCX_OPS_LOAD(__skel, __ops_name, __scx_name, __uei_name) ({		\
+	UEI_SET_SIZE(__skel, __ops_name, __uei_name);				\
+	if (__COMPAT_struct_has_field("sched_ext_ops", "exit_dump_len")) {	\
+		bpf_map__set_autocreate((__skel)->maps.__ops_name, true);	\
+		bpf_map__set_autocreate((__skel)->maps.__ops_name##___no_exit_dump_len, false); \
```
ByteLab-David, line 123: Hopefully down the road we'll be able to figure out a way to do this that doesn't require us to hard-code the BC struct-ops types like this. LGTM for now though

PR #166, opened 2024-03-28
Allow clean exits from the scheduler with scx_bpf_exit()

Byte-Lab:
There may be cases where a scheduler wants to cause itself to exit under
normal circumstances. For example, if a scheduler does not support
hotplug, it may want to exit to user space if a CPU is hotplugged or
unplugged to let it know that it should restart itself.

To enable this, we implement a new kfunc called scx_bpf_exit_bstr().
This kfunc shares a good amount of logic with scx_bpf_error_bstr(), as
both of them can plumb an exit message from BPF to user space. In
addition, we also augment struct scx_exit_info to include an exit code
field.

### DIFF
--- a/include/linux/sched/ext.h
+++ b/include/linux/sched/ext.h
@@ -56,7 +56,8 @@ enum scx_exit_kind {
 	SCX_EXIT_NONE,
 	SCX_EXIT_DONE,
 
-	SCX_EXIT_UNREG = 64,	/* BPF unregistration */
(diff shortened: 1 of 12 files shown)


ByteLab-David:
Spoke offline with Tejun. Going to change a few things around and send a v2

ByteLab-David:
Oops, missed fixing up scx_ops_exit_kind(). Will resend another patch shortly

ByteLab-David:
@htejun, this should be ready for another look

PR #174, opened 2024-04-05
scx: Cosmetic changes to kfuncs

htejun:
- Make __bpf_kfunc_start/end_defs() blocks strict so that they only contain kfunc definitions.

- Match BTF_KFUNCS table with kfunc definition blocks so that they always come right after in the same order.

- Add function comment to scx_bpf_select_cpu_dfl() and rename @found to @is_idle which is more accurate and consistent with the BPF side declaration.

### DIFF
--- a/kernel/sched/ext.c
+++ b/kernel/sched/ext.c
@@ -2141,21 +2141,6 @@ cpu_found:
 	return cpu;
 }
 
-__bpf_kfunc_start_defs();
-
-__bpf_kfunc s32 scx_bpf_select_cpu_dfl(struct task_struct *p, s32 prev_cpu,
-				       u64 wake_flags, bool *found)
-{
-	if (!scx_kf_allowed(SCX_KF_SELECT_CPU)) {
-		*found = false;
-		return prev_cpu;
-	}
-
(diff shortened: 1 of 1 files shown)


Byte-Lab (approved):

kernel/sched/ext.c
```diff
@@ -5153,17 +5149,62 @@ __bpf_kfunc void scx_bpf_error_bstr(char *fmt, unsigned long long *data,
...
+ * scx_bpf_put_idle_cpumask - Release a previously acquired referenced kptr to
+ * either the percpu, or SMT idle-tracking cpumask.
+ */
+__bpf_kfunc void scx_bpf_put_idle_cpumask(const struct cpumask *idle_mask)
+{
+	/*
+	 * Empty function body because we aren't actually acquiring or
+	 * releasing a reference to a global idle cpumask, which is read-only
+	 * in the caller and is never released. The acquire / release semantics
+	 * here are just used to make the cpumask is a trusted pointer in the
```
ByteLab-David, line 5205: Noticed a grammatical typo here from a previous commit. Feel free to fix, or ignore as you prefer.

htejun: Will fix. Thanks.

PR #177, opened 2024-04-09
scx: Exit a scheduler for unhandled hotplug events

Byte-Lab:
A scheduler may implement ops.cpu_online() and ops.cpu_offline() for CPU
onlining and offlining events respectively. If a scheduler does _not_
implement these callbacks, it's indicative that they do not support
hotplug. Given that a scheduler that doesn't support hotplug is
essentially guaranteed to perform incorrectly if a hotplug event occurs,
let's update ext.c to do the sane thing, and exit the scheduler
automatically.

Given that there may be other events in the future that could cause the
scheduler to initiate an exit, we create a new enum scx_exit_code type
that reserves the top bit of the exit_code field in struct
scx_exit_info, and defines SCX_ECODE_RESTART. We could as an alternative
instead just return something like -EAGAIN to signify that user space
can try restarting the scheduler.

Note that this isn't a 100% foolproof, race free hotplug detection mechanism.
For some schedulers, if a CPU is hotplugged after inspecting the host
topology but before attaching the scheduler, we could run into problems.
We'll need something like a hotplug generation counter to accommodate
this. We can take care of it in a separate follow up patch set.

### DIFF
--- a/include/linux/sched/ext.h
+++ b/include/linux/sched/ext.h
@@ -188,6 +188,27 @@ struct scx_cpu_release_args {
 	struct task_struct *task;
 };
 
+/*
+ * Exit code IDs are 64bit of the format:
+ *
+ *   Bits: [63  ..  48 47   ..  32 31 .. 0]
+ *         [ SYS ACT ] [ SYS RSN ] [ USR  ]
+ *
+ *   SYS ACT: System-defined exit actions
+ *   SYS RSN: System-defined exit reasons
+ *   USR    : User-defined exit codes and reasons
+ *
+ * Using the above mechanisms, users may communicate intention and context by
+ * ORing system actions and/or system reasons with a user-defined exit code.
+ */
+enum scx_exit_code {
+	/* Reasons */
+	SCX_ECODE_RSN_HOTPLUG	= 1LLU << 32,
+
+	/* Actions */
+	SCX_ECODE_ACT_RESTART	= 1LLU << 48,
+};
+
 /**
  * struct sched_ext_ops - Operation table for BPF scheduler implementation
  *
--- a/include/linux/sched/ext.h
+++ b/include/linux/sched/ext.h
@@ -58,6 +58,7 @@ enum scx_exit_kind {
 
 	SCX_EXIT_UNREG = 64,	/* User-space initiated unregistration */
 	SCX_EXIT_UNREG_BPF,	/* BPF-initiated unregistration */
+	SCX_EXIT_UNREG_KERN,	/* Main-kernel-initiated unregistration */
 	SCX_EXIT_SYSRQ,		/* requested by 'S' sysrq */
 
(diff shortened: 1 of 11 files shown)


arighi:
I'm not sure if relying purely on the implementation of `ops.cpu_online()` / `ops.cpu_offline()` is ideal. For instance, scx_simple works completely fine with cpu hotplugging, but it doesn't implement these methods. I'm wondering if having a dedicated flag would give more flexibility, something like `SCX_OPS_CPU_HOTPLUG_EXIT`` or similar? In this way it's up to the sched developer to determine if the scheduler supports cpu hotplugging correctly or not, wdyt?

ByteLab-David:
> I'm not sure if relying purely on the implementation of `ops.cpu_online()` / `ops.cpu_offline()` is ideal. For instance, scx_simple works completely fine with cpu hotplugging, but it doesn't implement these methods. I'm wondering if having a dedicated flag would give more flexibility, something like `SCX_OPS_CPU_HOTPLUG_EXIT`` or similar? In this way it's up to the sched developer to determine if the scheduler supports cpu hotplugging correctly or not, wdyt?

That would indeed be more convenient for `scx_simple`. On the other hand, `scx_simple` does have the flexibility of supporting hotplug by implementing empty versions of `ops.online()` and `ops.offline()`, albeit with a bit more boilerplate than just specifying an ops flag. It's arguably a bit more confusing / leaky (in terms of the API) to have behavior depend on the implementation of callbacks, but there is a precedent for that. See e.g. `ops.update_idle()` and `ops->flags & SCX_OPS_KEEP_BUILTIN_IDLE`. We'd also have to decide what the behavior should be for a scheduler that implements ops.cpu_online() / ops.cpu_offline() but also specifies that flag. I'm guessing we would want to reject loading the scheduler? At that point, the API is arguably no more leaky than just implementing the restart behavior if the callbacks aren't specified.

I'd expect that the cases where the scheduler cares about hotplug are when they have some view of the host topology; and generally speaking I'd expect that to be the norm for most production schedulers. I think the question is whether we want to do what we think is "the right thing" in terms of scheduler correctness for the common case (meaning assuming a scheduler that doesn't implement hotplug is probably not going to do the right thing in the event of a hotplug event), or to instead have a standalone flag that completely controls the behavior. Given that I think it's probably atypical for a scheduler to be completely agnostic to hotplug changes, my inclination is to err on the side of always exiting the scheduler if the callbacks aren't defined.

Wdyt?

arighi:
Hm... good point about implementing an empty `cpu_online` / `cpu_offline` to maintain the scheduling active, despite the additional boilerplate code it's probably the right thing to do, and it doesn't add extra complexity to the code. And, as you pointed out, there's also the `update_idle()` precedent, so it makes this approach more valid. I also agree that the default behavior should be "exit on a cpu hotplug event, if the scheduler ignores cpu hotplugging events".

That said, I agree with this approach, thanks for clarifying it.

htejun (approved):

PR #178, opened 2024-04-09
scx: Reorg code a bit and add possible/online cpumask helpers

htejun:


tools/sched_ext/scx_qmap.bpf.c
```diff
@@ -371,10 +371,68 @@ s32 BPF_STRUCT_OPS(qmap_init_task, struct task_struct *p,
...
+	s32 cpu;
+	char buf[128] = "", *p;
+	int idx;
+
+	possible = scx_bpf_get_possible_cpumask();
+	online = scx_bpf_get_online_cpumask();
+	if (!online) {
+		scx_bpf_error("failed to obtain cpumasks");
+		return;
+	}
```
ByteLab-David: Does this load? Shouldn't the verifier complain that you're missing a release for `possible`?

htejun: Hah, lemme test again.

htejun: Nope, it doesn't complain. I'll fix it up but we should ping bpf folks.

PR #179, opened 2024-04-11
scx: Add hotplug sequence number

Byte-Lab:
We currently have a possibly tricky race w.r.t. hotplug that schedulers
don't have a good way to account for. Once a scheduler has inspected a
host topology, if a hotplug event occurs before a scheduler is attached
and loaded, then the scheduler will have no way of knowing that its view
of the host topology is incorrect. Hotplug events _after_ this are fine,
as we'll either pass the events to the scheduler, or evict the scheduler
directly. But if a hotplug event happens between inspecting the host
topology and attaching the scheduler, we have a problem.

This series addresses this by adding the capability to track a
monotonically increasing sequence number across hotplug events.

### DIFF
--- a/kernel/sched/ext.c
+++ b/kernel/sched/ext.c
@@ -26,7 +26,7 @@ struct scx_exit_info {
 	/* %SCX_EXIT_* - broad category of the exit reason */
 	enum scx_exit_kind	kind;
 
-	/* exit code if gracefully exiting from BPF */
+	/* exit code if gracefully exiting */
 	s64			exit_code;
 
 	/* textual representation of the above */
@@ -6068,7 +6068,7 @@ static int __init scx_init(void)
 
 	scx_kset = kset_create_and_add("sched_ext", &scx_uevent_ops, kernel_kobj);
 	if (!scx_kset) {
-		pr_err("sched_ext: Failed to create /sys/sched_ext\n");
+		pr_err("sched_ext: Failed to create /sys/kernel/sched_ext\n");
 		return -ENOMEM;
 	}
 
--- a/kernel/sched/ext.c
+++ b/kernel/sched/ext.c
@@ -562,6 +562,15 @@ struct sched_ext_ops {
 	u32 exit_dump_len;
 
 	/**
+	 * hotplug_seq - A sequence number that may be set by the scheduler to
+	 * detect when a hotplug event has occurred during the loading process.
+	 * If 0, no detection occurs. Otherwise, the scheduler will fail to
+	 * load if the sequence number does not match @scx_hotplug_seq on the
+	 * enable path.
+	 */
+	u64 hotplug_seq;
+
+	/**
 	 * name - BPF scheduler's name
 	 *
 	 * Must be a non-zero valid BPF object name including only isalnum(),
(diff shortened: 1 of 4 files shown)


ByteLab-David:
Latest push was to update the `SCX_HOTPLUG_SQN()` macro to be a `static inline` function instead.

kernel/sched/ext.c
```diff
@@ -796,6 +796,7 @@ static atomic_t scx_exit_kind = ATOMIC_INIT(SCX_EXIT_DONE);
 static struct scx_exit_info *scx_exit_info;
 
 static atomic_long_t scx_nr_rejected = ATOMIC_LONG_INIT(0);
+static atomic_long_t scx_hotplug_sqn = ATOMIC_LONG_INIT(0);
```
htejun: `_seq` is probably more consistent as we already have `scx_kick_cpus_pnt_seqs`.

kernel/sched/ext.c
```diff
@@ -4346,6 +4355,28 @@ static struct kthread_worker *scx_create_rt_helper(const char *name)
 	return helper;
 }
 
+static int validate_ops_cpus_locked(const struct sched_ext_ops *ops)
```
htejun: Let's use a name which explicitly states that it's checking for hotplug. I'm a bit doubtful we'd add more ops verification at this stage.

kernel/sched/ext.c
```diff
@@ -4346,6 +4355,28 @@ static struct kthread_worker *scx_create_rt_helper(const char *name)
...
+	 * initialized, and when we were able to attach, exit and notify user
+	 * space about it.
+	 */
+	if (ops->hotplug_sqn) {
+		global_hotplug_sqn = atomic_long_read(&scx_hotplug_sqn);
+		if (ops->hotplug_sqn != global_hotplug_sqn) {
+			scx_ops_exit(SCX_ECODE_ACT_RESTART | SCX_ECODE_RSN_HOTPLUG,
+				     "expected hotplug sqn %llu did not match actual %llu",
+				     ops->hotplug_sqn, global_hotplug_sqn);
+			return -EBUSY;
```
htejun: Would -EAGAIN or sth else be better? We use -EBUSY to indicate that the scheduler is already loaded.

tools/testing/selftests/sched_ext/hotplug.c
```diff
@@ -93,12 +90,48 @@ static enum scx_test_status test_hotplug(bool onlining, bool cbs_defined)
...
+	/*
+	 * Take the CPU offline to increment the global hotplug sqn, which
+	 * should cause attach to fail due to us setting the hotplug sqn above
+	 */
+	toggle_online_status(0);
+	link = bpf_map__attach_struct_ops(skel->maps.hotplug_nocb_ops);
+
+	toggle_online_status(1);
+
+	SCX_ASSERT(!link);
```
htejun: So, this gives us two different modes that hotplug event is reported - attach failure and the SCX_EXITED() and then error code reporting. Wouldn't it be simpler if both were reported the same way? ie. Instead of failing attaching, we can just pretend that the hotplug event which raced attaching happened after attaching is complete.

kernel/sched/ext.c
ByteLab-David: Yeah that makes sense. This should however no longer apply now that we're purposefully using the same mechanism to communicate a hotplug error per your suggestion below.

tools/testing/selftests/sched_ext/hotplug.c
ByteLab-David: Good point -- functionally speaking they're the same thing, and this isn't a problem with the scheduler itself anyways so failing to attach may be confusing semantics as well.

I'll update the logic to do this in v2.

ByteLab-David:
Sorry, forgot to change the value in sched_ext_ops, 1 moment.

ByteLab-David:
This should be good to go

htejun (approved):

PR #180, opened 2024-04-12
scx: Implement cpufreq support

htejun:
Add `scx_bpf_cpuperf_cap()`, `scx_bpf_cpuperf_cur()` and scx_bpf_cpuperf_set()` which allows BPF schedulers to monitor and set each CPU's performance level. `ops.tick()` is added too.

### DIFF
--- a/kernel/sched/ext.c
+++ b/kernel/sched/ext.c
@@ -261,6 +261,16 @@ struct sched_ext_ops {
 	void (*dispatch)(s32 cpu, struct task_struct *prev);
 
 	/**
+	 * tick - Periodic tick
+	 * @p: task running currently
+	 *
+	 * This operation is called every 1/HZ seconds on CPUs which are
+	 * executing an SCX task. Setting @p->scx.slice to 0 will trigger an
+	 * immediate dispatch cycle on the CPU.
+	 */
+	void (*tick)(struct task_struct *p);
+
+	/**
 	 * runnable - A task is becoming runnable on its associated CPU
 	 * @p: task becoming runnable
 	 * @enq_flags: %SCX_ENQ_*
@@ -3097,6 +3107,8 @@ static void task_tick_scx(struct rq *rq, struct task_struct *curr, int queued)
 	if (scx_ops_bypassing()) {
 		curr->scx.slice = 0;
 		touch_core_sched(rq, curr);
+	} else if (SCX_HAS_OP(tick)) {
+		SCX_CALL_OP(SCX_KF_REST, tick, curr);
 	}
 
 	if (!curr->scx.slice)
--- a/kernel/sched/ext.c
+++ b/kernel/sched/ext.c
@@ -4474,7 +4474,7 @@ static int scx_ops_enable(struct sched_ext_ops *ops)
 	struct scx_task_iter sti;
 	struct task_struct *p;
 	unsigned long timeout;
-	int i, ret;
+	int i, cpu, ret;
 
 	mutex_lock(&scx_ops_enable_mutex);
 
@@ -4522,6 +4522,9 @@ static int scx_ops_enable(struct sched_ext_ops *ops)
 	scx_warned_zero_slice = false;
 
 	atomic_long_set(&scx_nr_rejected, 0);
+
+	for_each_possible_cpu(cpu)
+		cpu_rq(cpu)->scx.cpuperf_target = SCX_CPUPERF_ONE;
 
 	/*
 	 * Keep CPUs stable during enable so that the BPF scheduler can track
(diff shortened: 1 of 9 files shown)


tools/sched_ext/scx_qmap.bpf.c
```diff
@@ -432,12 +475,101 @@ void BPF_STRUCT_OPS(qmap_cpu_offline, s32 cpu)
...
+	u32 zero = 0;
+	u32 nr_cpu_ids = scx_bpf_nr_cpu_ids();
+	u64 cap_sum = 0, cur_sum = 0, cur_min = SCX_CPUPERF_ONE, cur_max = 0;
+	u64 target_sum = 0, target_min = SCX_CPUPERF_ONE, target_max = 0;
+	const struct cpumask *online;
+	int i, nr_online_cpus = 0;
+
+	online = scx_bpf_get_online_cpumask();
+	if (!online)
+		return -ENOMEM;
```
ByteLab-David, line 496: It shouldn't be necessary to check for nullity here -- the kfunc doesn't advertise that it can return `NULL` via `KF_RET_NULL`

kernel/sched/ext.c
```diff
@@ -16,6 +16,45 @@ enum scx_consts {
 	SCX_EXIT_BT_LEN			= 64,
 	SCX_EXIT_MSG_LEN		= 1024,
 	SCX_EXIT_DUMP_DFL_LEN		= 32768,
+
+	SCX_CPUPERF_ONE			= SCHED_CAPACITY_SCALE,
```
ByteLab-David, line 20: Could you please clarify the idea behind the name of this constant? Why wouldn't we want to call it `SCX_CPUPERF_MAX`? Is the idea that these relative capacity and frequency (or performance?) values we're querying are conceptually on a scale of `[0, 1]` relative to other cores on the system, even though in practice they're `[0, 1024]`?

kernel/sched/ext.c
```diff
@@ -223,6 +262,16 @@ struct sched_ext_ops {
...
 
+	/**
+	 * tick - Periodic tick
+	 * @p: task running currently
+	 *
+	 * This operation is called every 1/HZ seconds on CPUs which are
+	 * executing an SCX task. Setting @p->scx.slice to 0 will trigger an
+	 * immediate dispatch cycle on the CPU.
+	 */
+	void (*tick)(struct task_struct *p);
```
ByteLab-David, line 273: Do you think it would be useful to also pass the `s32 cpu`? Certainly not strictly necessary, but I'd imagine it will be pretty common for callers to use `scx_bpf_task_cpu()` and `bpf_get_smp_processor_id()`, so it might be nice for ergonomics.

kernel/sched/ext.c
```diff
@@ -5896,6 +5980,74 @@ __bpf_kfunc u32 scx_bpf_nr_cpu_ids(void)
...
+ * scx_bpf_cpuperf_set - Set the relative performance target of a CPU
+ * @cpu: CPU of interest
+ * @perf: target performance level [0, %SCX_CPUPERF_ONE]
+ * @flags: %SCX_CPUPERF_* flags
+ *
+ * Set the target performance level of @cpu to @perf. @perf is in linear
+ * relative scale between 0 and %SCX_CPUPERF_ONE. This determines how the
+ * schedutil cpufreq governor chooses the target frequency. The actual
+ * performance level chosen is dependent on the hardware and cpufreq driver in
+ * use and can be monitored using scx_bpf_cpuperf_cur().
```
ByteLab-David: Do you think it's worth mentioning that the latency for frequency changes to be observed is hardware dependent? Or should that already be pretty obvious to anyone who'd be playing with these knobs in the first place?

Byte-Lab (approved):
Great work, this looks good to me. I left a few comments, but overall this looks correct and like the right approach.

One thing I did want to ask -- now that we have support for dynamically setting performance levels for individual CPUs, do you think we should consider adding support for uclamp? Given that it's already an API that's available to users, and fits the schedutil CPU performance APIs quite well, it might be a natural API to expose and build on. Wdyt?

kernel/sched/ext.c
htejun: I think the benefit of keeping the interface params minimal likely outweighs the convenience factor. For internal functions, having extra args to avoid repeated lookups is fine but here I think it's better to keep things tighter.

kernel/sched/ext.c
htejun: Will fortify the comment a bit.

tools/sched_ext/scx_qmap.bpf.c
htejun: Ah, will drop. Thanks.

kernel/sched/ext.c
htejun: Yeah, it's treating it as fixed point scale where the actual number range is [0, 1]. In practice, it's the same as MAX but I tend to like ONEs better because it's conceptually more versatile. e.g. The current interface doesn't do it but if we want the interface turbo, doing it with ONE is more natural than with MAX.

htejun:
> One thing I did want to ask -- now that we have support for dynamically setting performance levels for individual CPUs, do you think we should consider adding support for uclamp? Given that it's already an API that's available to users, and fits the schedutil CPU performance APIs quite well, it might be a natural API to expose and build on. Wdyt?

uclamp overlays on top of schedutil and should work the same whether the underlying scheduler is fair or scx. ie. If the user clamps cpufreq range for a task, the clamping already works the same for scx.

PR #185, opened 2024-04-21
Implement BPF DSQ iterator

htejun:
This allows iterating tasks in a non-local DSQ in [reverse] dispatch order from any context. While iterating in `ops.dispatch()`, specific tasks can be consumed in any order too.


init/init_task.c
```diff
@@ -101,7 +101,7 @@ struct task_struct init_task __aligned(L1_CACHE_BYTES) = {
 #endif
 #ifdef CONFIG_SCHED_CLASS_EXT
 	.scx		= {
-		.dsq_node.fifo	= LIST_HEAD_INIT(init_task.scx.dsq_node.fifo),
+		.dsq_node.list	= LIST_HEAD_INIT(init_task.scx.dsq_node.list),
```
ByteLab-David, line 104: Should we also initialize `.dsq_node.flags`, `.dsq_node.priq`, and `.dsq_seq`?

kernel/sched/ext.c
```diff
@@ -5714,170 +5939,159 @@ __bpf_kfunc void scx_bpf_kick_cpu(s32 cpu, u64 flags)
...
 {
 	struct scx_dispatch_q *dsq;
+	s32 ret;
 
-	lockdep_assert(rcu_read_lock_any_held());
+	preempt_disable();
 
 	if (dsq_id == SCX_DSQ_LOCAL) {
-		return this_rq()->scx.local_dsq.nr;
+		ret = this_rq()->scx.local_dsq.nr;
```
ByteLab-David: Not your change, but given that we're reading `dsq->nr` without taking the dsq lock, do we need to read / write it with READ_ONCE() and WRITE_ONCE()?

init/init_task.c
htejun: I went the other direction and removed 0 inits. While there are some 0 inits for existing fields, it's far from being consistent. I think it'd be better to just init what's necessary.

include/linux/sched/ext.h
```diff
@@ -121,18 +121,20 @@ enum scx_kf_mask {
 	__SCX_KF_TERMINAL	= SCX_KF_ENQUEUE | SCX_KF_SELECT_CPU | SCX_KF_REST,
 };
 
+struct scx_dsq_node {
+	struct list_head	list;		/* dispatch order */
+	struct rb_node		priq;		/* p->scx.dsq_vtime order */
+	u32			flags;		/* SCX_TASK_DSQ_* flags */
```
ByteLab-David, line 130: `flags` I think we do actually have to initialize in `__sched_fork()`. It looks like we were missing `.dsq_flags` before. Also applies to `dsq_seq`

kernel/sched/ext.c
```diff
@@ -5857,6 +5894,111 @@ __bpf_kfunc void scx_bpf_destroy_dsq(u64 dsq_id)
...
+		return -EINVAL;
+
+	kit->dsq = find_non_local_dsq(dsq_id);
+	if (!kit->dsq)
+		return -ENOENT;
+
+	INIT_LIST_HEAD(&kit->cursor.list);
+	RB_CLEAR_NODE(&kit->cursor.priq);
+	kit->cursor.flags = SCX_TASK_DSQ_CURSOR;
+	kit->dsq_seq = kit->dsq->seq;
```
ByteLab-David: `READ_ONCE()`?

kernel/sched/ext.c
```diff
@@ -1498,6 +1530,11 @@ static void dispatch_enqueue(struct scx_dispatch_q *dsq, struct task_struct *p,
 		else
 			list_add_tail(&p->scx.dsq_node.list, &dsq->list);
 	}
+
+	/* seq records the order tasks are queued, used by BPF iterations */
+	dsq->seq++;
```
ByteLab-David, line 1541: I think we need a `WRITE_ONCE()` here as well

tools/sched_ext/scx_qmap.bpf.c
```diff
@@ -258,7 +260,7 @@ static void update_core_sched_head_seq(struct task_struct *p)
 void BPF_STRUCT_OPS(qmap_dispatch, s32 cpu, struct task_struct *prev)
 {
 	struct cpu_ctx *cpuc;
-	u32 zero = 0;
+	u32 zero = 0, batch = dsp_batch ?: 1;
```
ByteLab-David, line 300: Perhaps it'd be slightly clearer if we just initialized `dsp_batch` to 1?

kernel/sched/ext.c
```diff
@@ -5857,6 +5894,111 @@ __bpf_kfunc void scx_bpf_destroy_dsq(u64 dsq_id)
...
+		return;
+
+	if (!list_empty(&kit->cursor.list)) {
+		unsigned long flags;
+
+		raw_spin_lock_irqsave(&kit->dsq->lock, flags);
+		list_del_init(&kit->cursor.list);
+		raw_spin_unlock_irqrestore(&kit->dsq->lock, flags);
+	}
+	kit->dsq = NULL;
```
ByteLab-David, line 6134: Hmm, why is it necessary to set to `NULL` here? Can the destroy kfunc be invoked more than once? Or is it just for sanity / to set the iterator state to an expected pre-init value?

kernel/sched/ext.c
```diff
@@ -2160,10 +2135,23 @@ static bool consume_dispatch_q(struct rq *rq, struct rq_flags *rf,
...
 	raw_spin_unlock(&dsq->lock);
-	return true;
+}
 
-remote_rq:
 #ifdef CONFIG_SMP
+static bool task_can_run_on_remote_rq(struct task_struct *p, struct rq *rq)
+{
+	return likely(test_rq_online(rq)) && !is_migration_disabled(p) &&
+		cpumask_test_cpu(cpu_of(rq), p->cpus_ptr);
```
ByteLab-David: This seems fine, but perhaps we should consider factoring out `is_cpu_allowed()` from core.c?

kernel/sched/ext.c
```diff
@@ -5731,12 +5737,82 @@ __bpf_kfunc bool scx_bpf_consume(u64 dsq_id)
...
+	}
+
+	/*
+	 * @kit can't be trusted and we can only get the DSQ from @p. As we
+	 * don't know @p's rq is locked, use READ_ONCE() to access the field.
+	 * Derefing is safe as DSQs are RCU protected.
+	 */
+	dsq = READ_ONCE(p->scx.dsq);
+	if (!dsq || dsq != kit_dsq)
+		return false;
```
ByteLab-David: Do we need to verify that the `dsq` is non-local? What if, for example, a malicious BPF program passes some stack-allocated `struct bpf_iter_scx_dsq_kern` that points to a local DSQ to which `p` has already been dispatched? Could we accidentally try to consume from it?

tools/sched_ext/scx_qmap.bpf.c
```diff
@@ -260,6 +262,49 @@ static void update_core_sched_head_seq(struct task_struct *p)
...
+		return scx_bpf_consume(SHARED_DSQ);
+
+	/*
+	 * To demonstrate the use of scx_bpf_consume_task(), implement silly
+	 * selective priority boosting mechanism by scanning SHARED_DSQ looking
+	 * for matching comms and consume them first. This makes difference only
+	 * when dsp_batch is larger than 1.
+	 */
+	consumed = false;
+	bpf_for_each(scx_dsq, p, SHARED_DSQ, 0) {
```
ByteLab-David, line 281: Should we maybe add a macro to `common.bpf.h` so that users don't have to manually specify `scx_dsq` which is a bit leaky? Can be done in a follow-on change series, assuming you agree that it'd be useful.

tools/sched_ext/scx_qmap.bpf.c
```diff
@@ -260,6 +262,49 @@ static void update_core_sched_head_seq(struct task_struct *p)
...
+		memcpy(comm, p->comm, sizeof(exp_prefix) - 1);
+
+		bpf_for(i, 0, sizeof(exp_prefix)) {
+			if (exp_prefix[i] == '\0')
+				break;
+			if (exp_prefix[i] != comm[i]) {
+				match = false;
+				break;
+			}
+		}
```
ByteLab-David: `bpf_strncmp()` might be useful here

tools/sched_ext/scx_qmap.bpf.c
```diff
@@ -260,6 +262,49 @@ static void update_core_sched_head_seq(struct task_struct *p)
...
+		if (match && scx_bpf_consume_task(BPF_FOR_EACH_ITER, p)) {
+			consumed = true;
+			__sync_fetch_and_add(&nr_expedited, 1);
+		}
+	}
+
+	if (consumed)
+		return true;
+	else
+		return scx_bpf_consume(SHARED_DSQ);
```
ByteLab-David: Totally optional suggestion: `return consumed || scx_bpf_consume(SHARED_DSQ);`

Byte-Lab (commented):
Overall looks great, this will be super useful. Left a few comments

include/linux/sched/ext.h
htejun: I factored sched_ext_entity into a function which memsets the whole thing to 0 before initing non-zero fields.

kernel/sched/ext.c
htejun: I don't know exactly but other iterators do that and I saw `next()` being called after `destroy()` after breaking out of the loop. If I don't clear dsq, that can lead to oopses.

kernel/sched/ext.c
htejun: Hmm... yeah, it'd be useful to audit the conditions against `is_cpu_allowed()` but I'm not sure we'd be able to use the exact same conditions. We start saying no as soon as cpu hotplug starts, so the other tests should be fine. I think the only thing missing is `task_cpu_possible()` test. Will add that.

tools/sched_ext/scx_qmap.bpf.c
htejun: I thought about it but not sure. `bpf_for_each()` is the generic helper and we're using it to iterate a `scx_dsq`, so as long as we have documentation pointing to example usage, I think we're fine.

tools/sched_ext/scx_qmap.bpf.c
htejun: That doesn't take `p->comm` unfortunately. :(

kernel/sched/ext.c
```diff
@@ -5576,12 +5768,88 @@ __bpf_kfunc bool scx_bpf_consume(u64 dsq_id)
...
+	}
+
+	/*
+	 * @kit can't be trusted and we can only get the DSQ from @p. As we
+	 * don't know @p's rq is locked, use READ_ONCE() to access the field.
+	 * Derefing is safe as DSQs are RCU protected.
+	 */
+	dsq = READ_ONCE(p->scx.dsq);
+
+	if (unlikely(dsq->id == SCX_DSQ_LOCAL)) {
```
ByteLab-David, line 5807: optional suggestion: given that we don't support iterating over non-builtins, maybe this will cover all our bases a bit better?

```
if (unlikely(dsq->id & SCX_DSQ_BUILTIN))
```

Byte-Lab (approved):
LGTM, looks fantastic. Left one optional comment but this looks good to go

kernel/sched/ext.c
ByteLab-David: Chatted offline: global DSQ iteration is supported, so this is fine in its current form.

PR #197, opened 2024-05-03
scx: Dispatch locally until ops.cpu_online() invoked

Byte-Lab:
In sched_ext, there's currently a corner case on the hotplug path wherein a CPU may be onlined and begin scheduling _before_ we've invoked ops.cpu_online() to the scheduler. This can cause confusion by e.g. causing a scheduler to error out if it receives an ops.dispatch() call and accidentally dispatches to an invalid DSQ.

To avoid this scenario, this patch adds a per-CPU boolean variable that tracks when a CPU has been onlined from the perspective of sched_ext.

Note that this patch does not prevent _any_ callback from being invoked on that CPU. It simply prevents us from invoking ops.dispatch(), and forces us to dispatch directly to the local DSQ. We can still, however, receive e.g. ops.runnable() if a per-CPU task is dispatched to the local DSQ of the onlined CPU before the hotplug event is received.

### DIFF
--- a/kernel/sched/ext.c
+++ b/kernel/sched/ext.c
@@ -920,6 +920,16 @@ static unsigned long __percpu *scx_kick_cpus_pnt_seqs;
  */
 static DEFINE_PER_CPU(struct task_struct *, direct_dispatch_task);
 
+/*
+ * Has the current CPU been onlined from the perspective of scx?
+ *
+ * A hotplugged CPU may begin scheduling tasks before the core scheduler will
+ * call into rq_online_scx(). Track whether we've had a chance to invoke
+ * ops.cpu_online() so we can dispatch locally until the scheduler knows about
+ * the hotplug event.
+ */
+static DEFINE_PER_CPU(bool, scx_cpu_online);
+
 /* dispatch queues */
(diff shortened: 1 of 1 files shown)


kernel/sched/ext.c
```diff
@@ -3202,10 +3218,12 @@ static void rq_online_scx(struct rq *rq, enum rq_onoff_reason reason)
 {
 	if (reason == RQ_ONOFF_HOTPLUG)
 		handle_hotplug(rq, true);
+	__this_cpu_write(scx_cpu_online, 1);
```
htejun: true and false?

kernel/sched/ext.c
```diff
@@ -1773,6 +1783,11 @@ static void direct_dispatch(struct task_struct *p, u64 enq_flags)
 	dispatch_enqueue(dsq, p, enq_flags);
 }
 
+static bool scx_cpu_onlined(void)
```
htejun: It might be better to not have this helper. It's tested in only one place and the naming gets a bit crowded, I think.

htejun (approved):

PR #199, opened 2024-05-06
Dispatch hotplug

Byte-Lab:
In commit 57845b39e8bb ("scx: Dispatch locally until ops.cpu_online() invoked"), we updated ext.c to only issue dispatch callbacks if we know that a CPU has been fully onlined. There were a couple of small issues with this commit, including:

1. The documentation wasn't quite right -- it said that we choose to dispatch locally, but that was a stale comment.

2. We now have two places where we're checking CPU online / offline status, which is unnecessarily confusing. scx_cpu_online matches exactly where the hotplug callbacks are invoked, so we can just use those everywhere.

Note that we can still invoke callbacks for e.g. per-CPU tasks that run on onlining CPUs before the CPU has been fully onlined. We'll update our documenation in the next patch to clarify this.

### DIFF
--- a/kernel/sched/ext.c
+++ b/kernel/sched/ext.c
@@ -303,9 +303,10 @@ struct sched_ext_ops {
 	 *
 	 * This and ->enqueue() are related but not coupled. This operation
 	 * notifies @p's state transition and may not be followed by ->enqueue()
-	 * e.g. when @p is being dispatched to a remote CPU. Likewise, a task
-	 * may be ->enqueue()'d without being preceded by this operation e.g.
-	 * after exhausting its slice.
+	 * e.g. when @p is being dispatched to a remote CPU, or when @p is
+	 * being enqueued on a CPU experiencing a hotplug event. Likewise, a
+	 * task may be ->enqueue()'d without being preceded by this operation
+	 * e.g. after exhausting its slice.
 	 */
 	void (*runnable)(struct task_struct *p, u64 enq_flags);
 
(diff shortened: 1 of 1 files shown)


kernel/sched/ext.c
```diff
@@ -2572,7 +2573,7 @@ static int balance_one(struct rq *rq, struct task_struct *prev,
 		goto has_tasks;
 
 	if (!SCX_HAS_OP(dispatch) || scx_ops_bypassing() ||
-			unlikely(!*this_cpu_ptr(&scx_cpu_online)))
+			unlikely(!test_rq_online(rq)))
```
htejun: Can you indent the second line so that it aligns with the opening parenthesis?

kernel/sched/ext.c
```diff
@@ -1786,7 +1787,7 @@ static void direct_dispatch(struct task_struct *p, u64 enq_flags)
 static bool test_rq_online(struct rq *rq)
```
htejun, line 1787: Maybe `test_cpu_online()`?

kernel/sched/ext.c
```diff
@@ -1786,7 +1787,7 @@ static void direct_dispatch(struct task_struct *p, u64 enq_flags)
...
 {
 #ifdef CONFIG_SMP
```
htejun: As we aren't making `scx_cpu_online` conditional, we might as well just drop the ifdef.

htejun (approved):

kernel/sched/ext.c
htejun: Never mind. Let's leave the name as-is.

htejun (approved):

PR #200, opened 2024-05-06
scx: Fix exit/error bstr formatting

Byte-Lab:
In a recent commit, we refactored bpf_exit_bstr_common() into bstr_format(). This unfortunately regressed printing output passed from BPF, as we incorrectly invoked bpf_exit_bstr_common() in the caller. Fix it.


ByteLab-David:
I noticed this when running selftests locally, not sure why it didn't catch it in CI. I'll take a look

htejun (approved):

PR #201, opened 2024-05-08
Improve debug dump

htejun:
- Dumps can now be read through `sched_ext_dump` tracepoint.
- Dumps can now be triggered non-destructively using `sysrq-D`.

### DIFF
--- a/kernel/sched/ext.c
+++ b/kernel/sched/ext.c
@@ -963,6 +963,15 @@ struct scx_dsp_ctx {
 
 static DEFINE_PER_CPU(struct scx_dsp_ctx, scx_dsp_ctx);
 
+/* string formatting from BPF */
+struct scx_bstr_buf {
+	u64			data[MAX_BPRINTF_VARARGS];
+	char			line[SCX_EXIT_MSG_LEN];
+};
+
+static DEFINE_RAW_SPINLOCK(scx_bstr_buf_lock);
+static struct scx_bstr_buf scx_bstr_buf;
+
 /* ops debug dump */
 static s32 scx_dump_cpu = -1;
 static char scx_dump_last;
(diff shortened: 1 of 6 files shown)


kernel/sched/ext.c
```diff
@@ -4497,25 +4517,115 @@ static void scx_ops_disable(enum scx_exit_kind kind)
...
+static void dump_newline(struct seq_buf *s)
+{
+	trace_sched_ext_dump("");
+
+	/* @s may be zero sized and seq_buf triggers WARN if so */
+	if (s->size)
+		seq_buf_putc(s, '\n');
+}
+
+static __printf(2, 0) void dump_line(struct seq_buf *s, const char *fmt, ...)
```
ByteLab-David: Should this be `__printf(2, 3)`?

tools/sched_ext/include/scx/common.h
```diff
@@ -26,6 +26,21 @@ typedef int16_t s16;
 typedef int32_t s32;
 typedef int64_t s64;
 
+/*
+ * Let's avoid including whole vmlinux.h from userspace code by copying what's
+ * needed from vmlinux.h. This is error-prone. Hopefully, we'll figure out a
+ * better way to do this in the future.
+ */
+enum scx_ops_flags {
```
ByteLab-David: Per our offline discussion, we could use `__COMPAT_ENUM_OR_ZERO()` here to give this CO-RE semantics. Something like:
```
define SCX_OPS_KEEP_BUILTIN_IDLE __COMPAT_enum_or_zero("scx_ops_flags", SCX_OPS_KEEP_BUILTIN_IDLE)
```
Feel free to ignore or apply as you think best.

Byte-Lab (approved):
Looks great. Thanks a lot for making the debug story so much better with these last few change sets.

tools/sched_ext/include/scx/common.h
htejun: I dropped this patch. Let's do it in a separate patch. In an earlier version, I added a new ops flag for trace dumping but then switched to using a TP which makes this change unnecessary for this PR. I'll add this later.

kernel/sched/ext.c
htejun: Fixed.

PR #205, opened 2024-05-17
scx: Close a small race window in the enable path which can lead to use-after-free

htejun:
scx_ops_enable() was iterating all tasks without locking its rq and then calling get_task_struct() on them. However, this can race against task free path. The task struct itself is accessible but its refcnt may already have reached zero by the time get_task_struct() is called. This triggers the following refcnt warning.

  WARNING: CPU: 11 PID: 2834521 at lib/refcount.c:25 refcount_warn_saturate+0x7d/0xe0
  ...
  RIP: 0010:refcount_warn_saturate+0x7d/0xe0
  ...
  Call Trace:
   bpf_scx_reg+0x851/0x890
   bpf_struct_ops_link_create+0xbd/0xe0
   __sys_bpf+0x2862/0x2d20
   __x64_sys_bpf+0x18/0x20
   do_syscall_64+0x3d/0x90
   entry_SYSCALL_64_after_hwframe+0x46/0xb0

This can also affect other loops as it can call sched functions on tasks which are already destroyed. Let's fix it by making scx_task_iter_filtered_locked() also filter out TASK_DEAD tasks and switching the scx_task_iter_filtered() usage in scx_ops_enable() to scx_task_iter_filtered_locked(). This makes the next function always test TASK_DEAD while holding the task's rq lock and the rq lock guarantees that the task won't be freed closing the race window.

### DIFF
--- a/kernel/sched/ext.c
+++ b/kernel/sched/ext.c
@@ -1269,6 +1269,28 @@ static void scx_task_iter_init(struct scx_task_iter *iter)
 }
 
 /**
+ * scx_task_iter_rq_unlock - Unlock rq locked by a task iterator
+ * @iter: iterator to unlock rq for
+ *
+ * If @iter is in the middle of a locked iteration, it may be locking the rq of
+ * the task currently being visited. Unlock the rq if so. This function can be
+ * safely called anytime during an iteration.
+ *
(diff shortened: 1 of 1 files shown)


Byte-Lab (approved):
Nice debugging

PR #211, opened 2024-05-22
scx: Invoke ops.exit_task() for TASK_DEAD tasks on disable path

Byte-Lab:
In commit f26683180087 ("scx: Close a small race window in the enable path which can lead to use-after-free"), we fixed a race window on the enable path that could cause a crash. This fix is fine, but was a bit too aggressive in that it could also cause us to miss ops.exit_task() invocations in the following scenario:

1. A task exits and invokes do_task_dead() (making its state TASK_DEAD), but someone still holds a refcount on it somewhere.

2. The scheduler is disabled.

3. On the disable path, we don't invoke ops.task_exit()

4. We don't invoke it in sched_ext_free() either later, because by then the scheduler has been disabled.

Let's ensure we don't skip on exiting the task by still calling scx_ops_exit_task() for TASK_DEAD tasks on the disable path.


kernel/sched/ext.c
```diff
@@ -1333,14 +1333,16 @@ static struct task_struct *scx_task_iter_next(struct scx_task_iter *iter)
...
  *
- * Visit the next alive non-idle task with its rq lock held. See
- * scx_task_iter_init() for details.
+ * Visit the non-idle task with its rq lock held. Allows callers to specify
+ * whether they would like to filter out dead tasks. See scx_task_iter_init()
+ * for details.
  */
 static struct task_struct *
-scx_task_iter_next_filtered_locked(struct scx_task_iter *iter)
+scx_task_iter_next_locked(struct scx_task_iter *iter, bool filter)
```
htejun: Can we do `@include_dead` instead?

kernel/sched/ext.c
```diff
@@ -1367,12 +1369,38 @@ scx_task_iter_next_filtered_locked(struct scx_task_iter *iter)
...
+ * @iter: iterator to walk
+ *
+ * Visit the next non-idle task with its rq lock held. See scx_task_iter_init()
+ * for details.
+ */
+static struct task_struct *
+scx_task_iter_next_unfiltered_locked(struct scx_task_iter *iter)
+{
+	return scx_task_iter_next_locked(iter, false);
+}
```
htejun: I'm not sure what these wrapper add.

kernel/sched/ext.c
```diff
@@ -4393,19 +4421,22 @@ static void scx_ops_disable_workfn(struct kthread_work *work)
 
 	spin_lock_irq(&scx_tasks_lock);
 	scx_task_iter_init(&sti);
-	while ((p = scx_task_iter_next_filtered_locked(&sti))) {
+	while ((p = scx_task_iter_next_unfiltered_locked(&sti))) {
```
htejun: Given the subtlety, it probably warrants comment explaining this is done differently from other iterations?

Byte-Lab:
Sorry for so many force pushes, it should be good to go now.

ByteLab-David:
Sorry for so many force pushes, forgot to invert the polarity of the `include_dead` flag in the callers + check. It should be good to go now.

htejun (approved):

PR #215, opened 2024-05-31
scx: Fix CPU online tracking

Byte-Lab:
Commit 10d167d8ab81 ("SCX: Synchronize boot CPU's online state with SCX_RQ_ONLINE") fixes the oversight that we're not synchronizing the boot CPU's online state with scx_rq's online state. It does this by adding a check in init_sched_ext_class() for whether a CPU is online, and setting the SCX_RQ_ONLINE flag in the rq if it is.

The online mask may not be initialized by the time init_sched_ext_class() is invoked, so let's add a late init function where we iterate over the possible CPUs and do this check there.

### DIFF
--- a/kernel/sched/ext.c
+++ b/kernel/sched/ext.c
@@ -5741,9 +5741,6 @@ void __init init_sched_ext_class(void)
 		BUG_ON(!zalloc_cpumask_var(&rq->scx.cpus_to_preempt, GFP_KERNEL));
 		BUG_ON(!zalloc_cpumask_var(&rq->scx.cpus_to_wait, GFP_KERNEL));
 		init_irq_work(&rq->scx.kick_cpus_irq_work, kick_cpus_irq_workfn);
-
-		if (cpu_online(cpu))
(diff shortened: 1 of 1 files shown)


sirlucjan (approved):
The system under load behaved smoothly, no clipping, no loss of responsiveness. It looks like the bug reported by me and @ptr1337 has been fixed

ptr1337:
Can confirm, this patch fixes the sched-ext regression.

Tested-by: Peter Jung <ptr1337@cachyos.org>

sirlucjan:
Tested-by: Piotr Gorski [piotrgorski@cachyos.org](mailto:piotrgorski@cachyos.org)

ByteLab-David:
Hmm, multiple CachyOS users (using x86) are swearing that this PR improves interactivity over @htejun's original fix. I'm confused as to why, as it should have been a no-op. Will have to look more closely next week.

ByteLab-David:
I'm going to close this out -- there really shouldn't be any difference between top of tree and this fix. If someone is able to show something statistically significant we can reopen and revisit.

PR #217, opened 2024-06-05
scx: Fix hotplug

Byte-Lab:
The rq_online() and rq_offline() sched_class callbacks may be invoked on either
the domain creation path, or on the hotplug path. ext.c will only
invoke its hotplug logic when those callbacks are invoked on the hotplug path.
This turned out to be a bit brittle, as hotplug could break for sched_ext if we
first invoked rq_online() from the domain path, as it would cause us to skip
invoking it on the hotplug path due to hotplug implementation details. This in
fact happened after rebasing onto v6.10.

To avoid this frailty (and fix a hotplug regression for us after merging
v6.10), let's instead just call directly into ext.c on the hotplug path. This
could be made into a sched_class callback if that's what schedulers want, but
that's kind of what rq_online() and rq_offline() are supposed to be, so it's a
 bit awkward as is. It would have been ideal if we could have just used a
hotplug event notifier, but as far as I know that doesn't exist.


kernel/sched/core.c
```diff
@@ -9833,9 +9834,10 @@ int sched_cpu_deactivate(unsigned int cpu)
 	rq_lock_irqsave(rq, &rf);
 	if (rq->rd) {
 		BUG_ON(!cpumask_test_cpu(cpu, rq->rd->span));
-		set_rq_offline(rq, RQ_ONOFF_HOTPLUG);
+		set_rq_offline(rq);
 	}
 	rq_unlock_irqrestore(rq, &rf);
+	scx_rq_deactivate(rq);
```
htejun: Wouldn't it make more sense for `scx_rq_deactivate()` to be called before the locked block? To be symmetric with the online path?

htejun (commented):
Left one inline comment and I think the hotplug ops invocations would need to be updated to use `KF_SLEEPABLE`.

ByteLab-David:
>Left one inline comment and I think the hotplug ops invocations would need to be updated to use KF_SLEEPABLE.

Good, sorry for the oversight. I've added a call to a blockable kfunc in the testcase which would've noticed that.

htejun (approved):

